PR #13, opened 2015-11-25
modified app version string to be based on package version.

mandric:
Issue: https://github.com/medic/medic-webapp/issues/1664


alxndrsn:
Not sure why this PR has hung around so long - looks like you fixed this with https://github.com/medic/medic-android/commit/e499d4b0772cd786f4308b3da1239816e4246c7f :¬)

PR #19, opened 2016-03-08
Warn user when using a prod server with a dev app

alxndrsn:
Add an alarming red border if using configurable (i.e. dev) app with a medic production server.


SCdF:
:shipit:

PR #38, opened 2017-08-10
Refresh the lastKnownLocation every 5 min (for now)

derickl:
The current GPS location strategy is best-effort. The container currently doesn't
actively poll the GPS tranceiver but implicitly relies on other apps
actively using GPS to update the OS' lastKnownLocation.

We probe the locationManager for updates immediately it's set to get a head start in pulling that first gps lock.

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
@@ -4,8 +4,10 @@ import android.annotation.SuppressLint;
 import android.app.DatePickerDialog;
 import android.location.Criteria;
 import android.location.Location;
+import android.location.LocationListener;
 import android.location.LocationManager;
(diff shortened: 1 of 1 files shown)


alxndrsn:
@derickl thanks for this - I think it clearly identifies the problem and demonstrates some of the available methods for making sure that the location is reasonably up-to-date.  However, the call to `requestSingleUpdate()` is asynchronous, and doesn't guarantee an immediate response.  Also, `requestLocationUpdates()` will be triggered every time `getLocation()` is called, which will likely leak memory.

In the short term, I'll rework this PR to trigger `requestLocationUpdates()` at app startup; in the longer term I'll check:
* if newer devices will reliably support location requests directly through standard browser APIs (most preferable, as this will transparently support non-android clients)
* should we be supporting location requests asynchronously using `requestSingleUpdate()` in the client and a web-based handler for the callback?

PR #45, opened 2017-10-17
Handle photos on android 4.4 properly

alxndrsn:
On some handsets, we're provided with photos in the intent data extra instead of just being provided with the URI to the content of the photo.  In these cases, we need to write the supplied data to a file, and provide this file's URI to the webview.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -14,8 +14,10 @@
 	<uses-permission android:name="android.permission.VIBRATE"/>
 	<uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT"/>
 	<uses-permission android:name="com.android.launcher.permission.UNINSTALL_SHORTCUT"/>
-	<!-- READ_EXTERNAL_STORAGE required if users want to include photos from their phone -->
-	<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+	<!-- READ_EXTERNAL_STORAGE required if users want to include photos from their phone;
+	     WRITE_EXTERNAL_STORAGE required for taking new photos, and our resizing code.
+	     WRITE implies READ. -->
+	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 	<application android:label="@string/app_name"
 			android:icon="@mipmap/ic_launcher"
 			android:allowBackup="false"
--- a/src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
@@ -1,0 +1,200 @@
+package org.medicmobile.webapp.mobile;
+
+import android.app.Activity;
+import android.app.ProgressDialog;
+import android.content.Intent;
+import android.graphics.Bitmap;
+import android.graphics.Matrix;
+import android.net.Uri;
+import android.os.AsyncTask;
+import android.webkit.ValueCallback;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+import static android.app.Activity.RESULT_OK;
+import static android.provider.MediaStore.ACTION_IMAGE_CAPTURE;
+import static com.mvc.imagepicker.ImagePicker.getImageFromResult;
+import static com.mvc.imagepicker.ImagePicker.getPickImageIntent;
+import static org.medicmobile.webapp.mobile.EmbeddedBrowserActivity.GRAB_PHOTO;
+import static org.medicmobile.webapp.mobile.MedicLog.trace;
+import static org.medicmobile.webapp.mobile.MedicLog.warn;
+import static org.medicmobile.webapp.mobile.Utils.intentHandlerAvailableFor;
+import static org.medicmobile.webapp.mobile.Utils.showSpinner;
+
+class PhotoGrabber {
+	/** Max permitted file size for uploaded photos, in bytes. */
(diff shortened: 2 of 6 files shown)


alxndrsn:
@SCdF ready for review once more

src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
```diff
@@ -0,0 +1,203 @@
...
+	 *
+	 * @return the {@code android.net.Uri} of the created file
+	 */
+	private Uri writeToFile(Bitmap bitmap) throws IOException {
+		File temp = tempFile();
+		long tempFileSize = -1;
+		int quality;
+
+		do {
+			quality = 100;
```
SCdF: FWIW if this is JPEG quality levels I don't think you really ever want 100, it's kind of nuts for size vs quality[1]. I think 90 is a good starting quality, and almost certainly indistinguishable from 100.

[1] Stats from a random a 6000x4000 picture I have. 100 = 27.5mb, 90 = 9.2mb, 80 = 7mb, 70 = 5.2mb, etc.

src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,7 @@
 	<uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT"/>
 	<uses-permission android:name="com.android.launcher.permission.UNINSTALL_SHORTCUT"/>
 	<!-- READ_EXTERNAL_STORAGE required if users want to include photos from their phone -->
-	<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```
SCdF: You don't need both?

alxndrsn: No, `WRITE` implies `READ`.  Comment above looks wrong now, though :~)

src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
```diff
@@ -0,0 +1,203 @@
...
+	/** Smallest permitted image dimension in pixels */
+	private static final int MINIMUM_DIMENSION = 240;
+	/**
+	 * This is actually the value of
+	 * {@code ImagePicker.DEFAULT_REQUEST_CODE}.  This value is not exposed
+	 * by the library, but needs to be supplied to
+	 * {@code getImageFromResult()} so that it properly returns the full-
+	 * size image.
+	 */
+	private static final int FAKE_REQUEST_CODE = 234;
```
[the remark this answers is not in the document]
alxndrsn: Nope, we use `GRAB_PHOTO` as the request code.

src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	void process(int requestCode, int resultCode, Intent i) {
+		if(uploadCallback == null) {
+			warn(this, "uploadCallback is null for requestCode %s", requestCode);
+			return;
+		}
+
+		if(resultCode != RESULT_OK) {
+			// something went wrong; error code was logged above
```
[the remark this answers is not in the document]
alxndrsn: Err yeah, someone's deleted the logging that used to appear above 😉

src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
```diff
@@ -0,0 +1,203 @@
...
+		do {
+			quality = 100;
+
+			do {
+				trace(this, "writeToFile() :: Attempting to write bitmap=%s (%sx%s) to file=%s at quality=%s; size of last write=%s...",
+						bitmap, bitmap.getWidth(), bitmap.getHeight(), temp, quality, tempFileSize);
+
+				writeToFile(temp, bitmap, quality);
+
+				quality -= 10;
```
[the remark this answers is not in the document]
alxndrsn: 👍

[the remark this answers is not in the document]
alxndrsn: Definitely was weird, although less weird when originally written 🙂

src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
alxndrsn: Is there any point going as low as `0`?  or should we run through `90`-`50` or something?

alxndrsn:
Merged manually.

src/main/java/org/medicmobile/webapp/mobile/PhotoGrabber.java
SCdF: Depends on the library, might be worth playing with. In any case I don't think 0 is actually valid (in imagemagick that means "default" I think, and then it guesses).

Unfortunately this stuff really depends on the situation that it's used in. I wouldn't store photos of my vacation with a jpeg quality of 30, but from how I understand it would be used, I think 30 is probably a good minimum.

For laughs, here is an image (scaled down to 8mpixels) with quality 1:

q01

And here it is at 30:

q30

And here it is at 70:

q70

The image is kind of blurry, but then if we're getting into pixel peeping to see differences then I think our interest in quality has progressed further than our use case :-)

alxndrsn: Interesting comparison!  I've set the quality range from 90 down to 50, and after that the dimensions will both be halved, and quality set back to 90.  I'll change minimum quality to 30.

PR #54, opened 2018-06-28
GPS check before requesting location updates from GPS

rosteve:
With regards to [this ticket](https://github.com/medic/medic-projects/issues/4317) , @Mercyamulele   I and @ngamita (from communication) have experienced some phones which are crashing after accepting the fine location permission and opening the app in the first launch and subsequent times. Most of these devices are said to be Tecno phones with version 6.0 and above.
 
I was able to get hold of one phone which was experience this and these are the phone specs of the sample phone I was testing.

# a sample device experiencing this

key | value(s)
---|---
device make | _Tecno_
device model | _W2_
android version | _6.0_
can the problem be recreated on this device? | _yes_

Yesterday at night I tested the phone after running this fix and the phone was no longer crashing.


alxndrsn:
Merged!  Will be released in `v0.4.10`

PR #63, opened 2018-11-06
Only allow webview to access expected URLs

alxndrsn:
#66

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,4 +39,4 @@ test:
 	${GRADLEW} androidCheck lintUnbrandedDebug test
 
 travis: test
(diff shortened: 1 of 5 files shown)


alxndrsn:
No, performance hasn't been checked.

kennsippell:
I think this is better handled through for both mobile and chrome users via a Content-Security-Policy header.

garethbowen:
@kennsippell We have a CSP but it's very weak - I've raised [an issue](https://github.com/medic/medic-webapp/issues/5082) to make it stricter. However I don't think any CSP config would stop linking to another site would it? So maybe both this change and a stricter CSP are useful?

PR #70, opened 2019-05-22
Navigate to root instead of the _rewrite path (after webapp 3.5)

kennsippell:
#94

WebApp v3.5 uses the domain's root url instead of the `/medic/_design/medic/_rewrite/` path. The app directly navigates to this rewrite path, which then redirects to root. This adds an extra layer of complexity to the cache and an unnecessary redirect. This change introduces a build-time switch `USE_URL_ROOT_PATH` which uses the root path instead of the rewrite path when set.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,6 +78,7 @@ android {
 
 	applicationVariants.all {
 		buildConfigField "boolean", "DISABLE_APP_URL_VALIDATION", "Boolean.parseBoolean(\"${System.env.DISABLE_APP_URL_VALIDATION}\")";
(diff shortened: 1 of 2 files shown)


garethbowen:
@kennsippell Any update here?

kennsippell:
@garethbowen  I need to dive a litter deeper into how this config works in the app to get this working. But I think it is a good change to make - just fell off my radar. Should we get an issue tracking it and prioritized?

garethbowen:
Yes I think so. Thanks.

kennsippell:
https://github.com/medic/medic-android/issues/94

kennsippell:
This PR has great intention but poor execution.

Issue #72, opened 2019-05-29
gradle sync fails because minSDKversion is in the manifest file not the build.gradle file.


newtewt:
Trying to run gradle sync is failing beceause were we have the minSDKVersion defined. This can be updated to the build.gradle defaultConfig per the warning. Looks like targetSDKversion is also generating a warning. We should update both to fix. 

```
ERROR: The minSdk version should not be declared in the android manifest file. You can move the version from the manifest to the defaultConfig in the build.gradle file.
Move minSdkVersion to build file and sync project
Affected Modules: medic-android

WARNING: The targetSdk version should not be declared in the android manifest file. You can move the version from the manifest to the defaultConfig in the build.gradle file.
Move targetSdkVersion to build file and sync project
Affected Modules: medic-android 
```

PR #73, opened 2019-05-29
fixes gradle sync error and warning

newtewt:
Moving from AndroidManifest.xml to build.gradle to fix an error(minSdkVersion) and warning(targetSdkVersion) for gradle sync.  #72

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,7 @@ android {
 		versionCode getVersionCode()
 		versionName getVersionName()
 		archivesBaseName = "${project.name}-${versionName}"
+		minSdkVersion 19
 	}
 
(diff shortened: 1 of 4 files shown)


newtewt:
@garethbowen it's failing locally on master as well. I'm not 100% sure why but I'll see if I can clean it up.

newtewt:
Ahh there were two issues failing the build. 

Linting config changed because we are setting targetSDKVersion in a different area now and we aren't building against latest sdk. 

From what I've read android studio 3.0 will auto generate an adaptive icon folder `mipmap-anydpi-v26` and was causing a linking error. @henokgetachew is this required? If it is we need the resources added to fix the linking error.

PR #75, opened 2019-06-11
Updated dependency

abbyad:
For https://github.com/medic/medic-android/issues/74


abbyad:
Although the build progresses with this change, we are not able to build the apps. I get the following error for `gradle assembleRelease`:

> Failed to obtain compression information for entry

Perhaps the change https://github.com/noveogroup/android-check/commit/1f87548c0686d823e5f1490a2053b70bd9f5022e is needed after all!

abbyad:
Merged given that it builds properly on Travis.

PR #77, opened 2019-06-13
updated moh zanzibar app names

kitsao:
Updated as follows
1. production app name to `Jamii ni Afya`
2. training app name to `Jamii ni Afya - Darasa`


abbyad:
For https://github.com/medic/medic-projects/issues/6352 and https://github.com/medic/medic-projects/issues/6353

Issue #78, opened 2019-06-14
After merge to master: build, version, sign, and upload through Travis


kennsippell:
Creating a public ticket:
https://github.com/medic/medic-projects/issues/2319
https://github.com/medic/medic-projects/issues/3186

I've spent the last few days tinkering. This is a short update with progress, remaining obstacles, and some questions about the approach.

Currently, I have a travis config which builds and uploads release-ready APKs for all flavors. It triggers for new tags in `master` which match the semver schema (`a.b.c`). Builds are optimized, versioned based on the tag, and signed. Should be release-ready, but haven't done any e2e testing.

**Build Times**
The build completes in about ~40-45 minutes. Travis has a timeout of 50 minutes. I've never seen it timeout yet, but we're walking pretty close to the line and will likely cross it if we add another flavor. I'm going to try to create two jobs which both build ~half of the flavors. Or I'm going to whitelist the flavors we actually want to build and release. Seems like ~5 minutes of setup/install, about ~30 minutes to  build, and ~10 minutes (its about 500MB) to upload to GitHub. 

*Warning When we Upload our APK to Google's App Console*
```
Unoptimized APK
Warning:

This APK results in unused code and resources being sent to users. Your app could be smaller if you used the Android App Bundle. By not optimizing your app for device configurations, your app is larger to download and install on users' devices than it needs to be. Larger apps see lower install success rates and take up storage on users' devices.

Resolution:
Use the Android App Bundle to automatically optimize for device configurations, or manage it yourself with multiple APKs.
```

This appears to be some fallout from our change to both `armeabi-v7a` and `arm64-v8a`.  So I've spent some time trying to build bundles instead of APKs. I've got this building, signed, but not quite deploying (one sticky bug remaining). But in Google's App Console I'm getting this error when I try to deploy these bundles: `To upload an Android App Bundle you must be enrolled in App Signing by Google Play.`. I'm unclear if we want to try building multiple APKs, or if we want to try bundling with Google's App Signing. Seems that Google heavily recommends the use of App Signing, but I'm unclear on how that would impact our side loading. @henokgetachew?

**Tests**
I have disabled tests at the moment, but will likely need a new travis job to get that going again also.

**Collect and Gateway**
I'll reproduce this process for `medic-collect` and `medic-gateway` after I completed a PR for this one. I'm wondering if we need x64 builds for those projects also?

**Automated Deploys to App Store**
This is a can of worms I have not yet opened.

# What I'm planning to deliver
[...]

PR #81, opened 2019-06-18
Adds processing of incoming field timeTaken from rdt app

wtekeu:
https://github.com/medic/medic/issues/5695

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/MrdtSupport.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/MrdtSupport.java
@@ -38,7 +38,13 @@ class MrdtSupport {
 					byte[] data = i.getByteArrayExtra("data");
 					String base64data = Base64.encodeToString(data, Base64.NO_WRAP);
 					long timeTaken = i.getLongExtra("timeTaken", 0);
-					return safeFormat("const api = angular.element(document.body).injector().get('AndroidApi'); try { api.v1.mrdtTimeTakenResponse('\"%s\"'); } catch(ex) {} api.v1.mrdtResponse('\"%s\"')", String.valueOf(timeTaken), base64data);
+
+					String javaScript = "const api = angular.element(document.body).injector().get('AndroidApi');"
+														+ "if (api.v1.mrdtTimeTakenResponse) {"
+														+ "	api.v1.mrdtTimeTakenResponse('\"%s\"');"
+														+ "}"
+														+ "api.v1.mrdtResponse('\"%s\"');";
+					return safeFormat(javaScript, String.valueOf(timeTaken), base64data);
 				} catch(Exception /*| JSONException*/ ex) {
 					warn(ex, "Problem serialising mrdt image.");
 					return safeFormat("console.log('Problem serialising mrdt image: %s')", ex);
--- a/src/main/java/org/medicmobile/webapp/mobile/MrdtSupport.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/MrdtSupport.java
@@ -39,7 +39,7 @@ class MrdtSupport {
 					String base64data = Base64.encodeToString(data, Base64.NO_WRAP);
 					long timeTaken = i.getLongExtra("timeTaken", 0);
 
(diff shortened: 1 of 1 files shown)


kennsippell:
The tests here are failing because of my work on #78. Unfortunately, Travis allows only one encrypted file per repository. I've broken the build temporarily to build/test the automated publishing of builds to playstore. I spoke with @wtekeu and I can unblock this either fast (revert my change, halt work on #78) or slow (finish #78 and merge).  For now, we've decided to that I'll push ahead and drive toward a quick merge.

kennsippell:
@wtekeu If you rebase this, you should see tests passing. Sorry for the break!

src/main/java/org/medicmobile/webapp/mobile/MrdtSupport.java
```diff
@@ -37,7 +37,8 @@ String process(int requestCode, int resultCode, Intent i) {
 				try {
 					byte[] data = i.getByteArrayExtra("data");
 					String base64data = Base64.encodeToString(data, Base64.NO_WRAP);
-					return safeFormat("angular.element(document.body).injector().get('AndroidApi').v1.mrdtResponse('\"%s\"')", base64data);
+					long timeTaken = i.getLongExtra("timeTaken", 0);
+					return safeFormat("angular.element(document.body).injector().get('AndroidApi').v1.mrdtResponse('\"%s|%s\"')", String.valueOf(timeTaken), base64data);
```
kennsippell: Can we use a `get('AndroidApi').v2` to make this backward compatible?

wtekeu: A v2 would not help I think cause the bulk of the logic is in the enketo widget .. I have implemented what I believe is a simple change to ensure that 3.6 works with the current version of the wrapper. I needed to swap the positions of timeTaken and data (image) to ensure that it works

kennsippell: A project running a version of the webapp < 3.6 would break mRDT if they upgrade to the new container. I think the change in the container needs to be made backward compatible, or it will really cause issues.

Have you thought about passing this data separately instead of changing the data schema that is passed to `mrdtResponse`? You could make the same call to `mrdtResponse` (not breaking its backward compatibility) and also make a new call to `mrdtTimestampResponse` or alike.

src/main/java/org/medicmobile/webapp/mobile/MrdtSupport.java
```diff
@@ -37,7 +37,8 @@ String process(int requestCode, int resultCode, Intent i) {
 				try {
 					byte[] data = i.getByteArrayExtra("data");
 					String base64data = Base64.encodeToString(data, Base64.NO_WRAP);
-					return safeFormat("angular.element(document.body).injector().get('AndroidApi').v1.mrdtResponse('\"%s\"')", base64data);
+					long timeTaken = i.getLongExtra("timeTaken", 0);
+					return safeFormat("const api = angular.element(document.body).injector().get('AndroidApi'); try { api.v1.mrdtTimeTakenResponse('\"%s\"'); } catch(ex) {} api.v1.mrdtResponse('\"%s\"')", String.valueOf(timeTaken), base64data);
```
kennsippell: Can you just check if `api.v1.mrdtTimeTakenResponse` is defined? Instead of doing the try/catch?

kennsippell: Consider making this a multi-line string for readability.

kennsippell: Should it be ```
api.v1.mrdtTimeTakenResponse('\"%s\"', String.valueOf(timeTaken));
api.v1.mrdtResponse('\"%s\"')", base64data);
```? 

Looks like you're passing both values into `mrdtResponse`

wtekeu: that's not correct .. `%s` is the placeholder and I have two of them .. and it was tested ..

PR #82, opened 2019-06-18
#78 - Automatically publish builds with tags to play store

kennsippell:
This change uses a free library called [fastlane](https://fastlane.tools/) to kick off gradle and upload the output to the Play Store in the alpha channel. 
#78 

Sample output for this change is available [here](https://github.com/medic/medic-android/releases/tag/v0.4.27). Build version v0.4.27 has been uploaded to the App Play Console as an alpha build for all flavors. Build version v0.4.27 has been released as beta for `MOH Zanzibar` as part of medic/medic-projects#6353.

A few notes on the code: 

* The android build is broken until this merges. Explained [here](https://github.com/medic/medic-android/pull/81#issuecomment-503239309).
* The build tasks now run in parallel and now takes ~23 minutes with tests (previously ~45 minutes).
* The mechanism for publishing to GitHub Releases remains unchanged. However, this scenario is supported by fastlane.


.travis.yml
```diff
@@ -10,13 +10,53 @@ android:
...
+    - script:
+      - fastlane deploy flavor:livinggoods
+      - fastlane deploy flavor:livinggoodskenya
+      - fastlane deploy flavor:livinggoods_assisted_networks
+    - script:
+      - fastlane deploy flavor:livinggoods_innovation_ke
+      - fastlane deploy flavor:livinggoods_innovation_ke_supervisor 
+      - fastlane deploy flavor:livinggoods_innovation_ke_hivst
+    - script:
+      # - fastlane deploy flavor:moh_kenya_siaya_white
```
garethbowen: Not white?

.travis.yml
```diff
@@ -10,13 +10,53 @@ android:
...
+      - fastlane deploy flavor:moh_kenya_siaya_black
+      - fastlane deploy flavor:moh_zanzibar_training
+      - fastlane deploy flavor:moh_zanzibar
+    - script:
+      - fastlane deploy flavor:musomali
+      - fastlane deploy flavor:pih_malawi
+      - fastlane deploy flavor:pih_malawi_supervisor
+    - script:
+      - fastlane deploy flavor:simprints
+      - fastlane deploy flavor:vhw_burundi
```
garethbowen: It'd be awesome to generate this list dynamically but I can't see any way to do it in yaml. We could write a script to generate it precommit or something but that's probably a bit messy. This is good for now.

README.md
```diff
@@ -26,19 +26,14 @@ To build and deploy APKs for all configured brands:
...
+Create a git tag starting with `v`, e.g. `v1.2.3` and push the tag to GitHub. 
 
-1. Create a git tag starting with `v`, e.g. `v1.2.3` and push the tag to GitHub. Travis CI will build the release and push a signed, versioned, release-ready bundle to https://github.com/medic/medic-android/releases.
-
-Then, for each brand you'd like to publish:
-
-1. Download all APKs for the flavor you'd like to release from [GitHub Releases](https://github.com/medic/medic-android/releases).
-1. Navigate to the Google Play Console. `Release management` > `App releases`.
-1. Upload all APKs
+Creating this tag will trigger a Travis CI to build, sign, and properly version the build. The release-ready APKs are available for side-loading from [GitHub Releases](https://github.com/medic/medic-android/releases) and are uploaded to the Google Play Console in the "alpha" channel. To release to the public, click "Release to Production" or "Release to Beta" via the Google Play Console for each flavor.
```
garethbowen: SWEET!

fastlane/Fastfile
```diff
@@ -0,0 +1,33 @@
...
+  lane :deploy do |options|
+    gradleTaskName = "assemble#{options[:flavor]}Release"
+    version = ENV['TRAVIS_TAG'].empty? ? 'SNAPSHOT' : ENV['TRAVIS_TAG']
+    package_name = options[:flavor] == 'unbranded' ? "org.medicmobile.webapp.mobile" : "org.medicmobile.webapp.mobile.#{options[:flavor]}"
+
+    gradle(task: gradleTaskName)
+
+    supply(
+      package_name: package_name,
+      track: "alpha",
```
garethbowen: I wonder if we should go straight to beta at this point? I mean, it's only run when tagging so presumably it's reasonably stable. I think of alpha as more like nightly builds.

Not a big deal either way, we can define "alpha" however we like.

fastlane/Fastfile
```diff
@@ -0,0 +1,33 @@
...
+      skip_upload_aab: true,
+      skip_upload_metadata: true,
+      skip_upload_images: true,
+      skip_upload_screenshots: true,
+      validate_only: false,
+
+      timeout: 3600,
+    )
+  end
+end
```
garethbowen: I see fastlane also does signing - is it worth using it for that too?

.travis.yml
kennsippell: It isn't actually on the play store yet. Unsure of the story, but we are building it. I've messaged the Siaya team and if we don't need it anymore I'll remove it.

.travis.yml
kennsippell: Thought about that also. If it were 50 or if the list changed frequently, I'd have done it. But since it is so short ...

fastlane/Fastfile
kennsippell: From what I've seen, a lot of the apps use the "beta" channel as "production" and they have no active production versions. This is why I elected to go with alpha.

fastlane/Fastfile
kennsippell: Saw that ... same for deploying to GitHub releases...  I recon a solid argument could be made for consolidating as much build logic as possible in a single system. Would certainly improve readability. 

I thought it was reasonable to keep it in gradle just because a) it's done and vetted and tested, b) this is my first experience with fastlane and i wasn't really ready to bet on it as the single point of consolidation, and c) keeping it in gradle allows us to keep a single, self-installing tool as the only requirement for the dev workflow.

PR #90, opened 2019-08-17
Add MoH Mali flavor

derickl:


.travis.yml
```diff
@@ -52,6 +52,7 @@ jobs:
     - script:
       - fastlane deploy flavor:simprints
       - fastlane deploy flavor:vhw_burundi
+      - fastlane deploy flavor:moh_mali
```
kennsippell: I was trying to sort these alphabetically, just so it's easier to confirm all falvors are being built. Would be great if you'd find time to update

PR #95, opened 2019-10-22
Expose device info to webapp via bridge

wtekeu:
Expose device info to webapp via bridge

Example of info shared by code:
```json
{
  "androidVersion":"7.0",
  "osApiLevel":24,
  "osVersion":"3.18.35(1560398223)",
  "device":"Studio_Mega",
  "model":"Studio Mega",
  "manufacturer":"BLU",
  "harware":"mt6580",
  "freeMemorySize":"1,550MB",
  "totalMemorySize":"4,308MB",
  "freeRAMSize":"231MB",
  "totalRAMSize":"894MB",
  "thresholdRAM":"127MB",
  "cpuInfo":"ARMv7 Processor rev 3 (v7l)"
}
```

https://github.com/medic/medic-android/issues/64

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
@@ -7,6 +7,9 @@ import android.app.DatePickerDialog;
 import android.location.Criteria;
 import android.location.Location;
 import android.location.LocationManager;
+import android.net.ConnectivityManager;
+import android.net.NetworkCapabilities;
+import android.net.NetworkInfo;
 import android.net.TrafficStats;
 import android.os.Process;
 import android.widget.DatePicker;
@@ -46,6 +49,7 @@ public class MedicAndroidJavascript {
 
 	private LocationManager locationManager;
 	private ActivityManager activityManager;
+	private ConnectivityManager connectivityManager;
 	private Alert soundAlert;
 
 	public MedicAndroidJavascript(EmbeddedBrowserActivity parent) {
@@ -65,6 +69,10 @@ public class MedicAndroidJavascript {
 
 	public void setActivityManager(ActivityManager activityManager) {
 		this.activityManager = activityManager;
+	}
+
+	public void setConnectivityManager(ConnectivityManager connectivityManager) {
+		this.connectivityManager = connectivityManager;
 	}
 
 //> JavascriptInterface METHODS
@@ -251,16 +259,40 @@ public class MedicAndroidJavascript {
 	public String getDeviceInfo() {
 		try {
 			if (activityManager == null) {
-				return jsonError("ActivityManager not set.  Cannot retrieve RAM info.");
+				return jsonError("ActivityManager not set. Cannot retrieve RAM info.");
 			}
+
+			if (connectivityManager == null) {
+				return jsonError("ConnectivityManager not set. Cannot retrieve network info.");
+			}
+
+			String versionName = parent.getPackageManager()
+					.getPackageInfo(parent.getPackageName(), 0)
+					.versionName;
+			JSONObject appObject = new JSONObject();
+			appObject.put("version", versionName);
 
 			String androidVersion = Build.VERSION.RELEASE;
 			int osApiLevel = Build.VERSION.SDK_INT;
 			String osVersion = System.getProperty("os.version") + "(" + android.os.Build.VERSION.INCREMENTAL + ")";
+			JSONObject softwareObject = new JSONObject();
+			softwareObject
+					.put("androidVersion", androidVersion)
+					.put("osApiLevel", osApiLevel)
+					.put("osVersion", osVersion);
+
 			String device = Build.DEVICE;
 			String model = Build.MODEL;
 			String manufacturer = Build.BRAND;
-			String harware = Build.HARDWARE;
+			String hardware = Build.HARDWARE;
+			Map<String, String> cpuInfo = getCPUInfo();
+			JSONObject hardwareObject = new JSONObject();
+			hardwareObject
+					.put("device", device)
+					.put("model", model)
+					.put("manufacturer", manufacturer)
+					.put("hardware", hardware)
+					.put("cpuInfo", new JSONObject(cpuInfo));
 
 			File dataDirectory = Environment.getDataDirectory();
 			StatFs dataDirectoryStat = new StatFs(dataDirectory.getPath());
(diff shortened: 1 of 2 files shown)


kennsippell:
Would this be a good time to include these memory attributes also?

Function | Result | Summary of Meaning
-- | -- | --
[Runtime.maxMemory](https://developer.android.com/reference/java/lang/Runtime#maxMemory()) | 134 MB | Returns the maximum amount of memory that the Java virtual machine will attempt to use.
ActivityManager.getMemoryClass | 128 MB | This gives you an idea of how hard a memory limit you should impose on your application to let the overall system work best.
[ActivityManager.getLargeMemoryClass](https://developer.android.com/reference/android/app/ActivityManager.html#getLargeMemoryClass()) | 256 MB | Return the approximate per-application memory class of the current device when an application is running with a large heap.
[MemoryInfo.threshold](https://developer.android.com/reference/android/app/ActivityManager.MemoryInfo#threshold) | 144 MB | The threshold of availMem at which [android] considers memory to be low and start killing background services and other non-extraneous processes.

https://github.com/medic/cht-core/issues/5540#issuecomment-516708745

One other suggestion from that thread was to:

> Capture any Android onLowMemory and onTrimMemory(int) events

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -230,6 +246,58 @@ public void sms_send(String id, String destination, String content) throws Excep
...
+			String manufacturer = Build.BRAND;
+			String harware = Build.HARDWARE;
+
+			File dataDirectory = Environment.getDataDirectory();
+			StatFs dataDirectoryStat = new StatFs(dataDirectory.getPath());
+			long dataDirectoryBlockSize = dataDirectoryStat.getBlockSizeLong();
+			long dataDirectoryAvailableBlocks = dataDirectoryStat.getAvailableBlocksLong();
+			long dataDirectoryTotalBlocks = dataDirectoryStat.getBlockCountLong();
+			String freeMemorySize = formatSize(dataDirectoryAvailableBlocks * dataDirectoryBlockSize);
+			String totalMemorySize = formatSize(dataDirectoryTotalBlocks * dataDirectoryBlockSize);
```
kennsippell: Can we also add Runtime.maxMemory? This returns the maximum amount of memory that the Java virtual machine will attempt to use.

wtekeu:
New json:
```json
{
  "app":{
    "version":"SNAPSHOT"
  },
  "software":{
    "androidVersion":"7.0",
    "osApiLevel":24,
    "osVersion":"3.18.35(1560398223)"
  },
  "hardware":{
    "device":"Studio_Mega",
    "model":"Studio Mega",
    "manufacturer":"BLU",
    "hardware":"mt6580",
    "cpuInfo":{
      "cores":"3",
      "model name":"ARMv7 Processor rev 3 (v7l)"
    }
  },
  "memory":{
    "free":"494MB",
    "total":"4,308MB"
  },
  "ram":{
    "free":"252MB",
    "total":"894MB",
    "threshold":"127MB",
    "maxMemory":"256MB"
  },
  "network":{
    "downSpeed":"1Gbps",
    "upSpeed":"1Gbps"
  }
}
```

wtekeu:
New json:
```json
{
  "app": {
    "version": "SNAPSHOT"
  },
  "software": {
    "androidVersion": "7.0",
    "osApiLevel": 24,
    "osVersion": "3.18.35(1560398223)"
  },
  "hardware": {
    "device": "Studio_Mega",
    "model": "Studio Mega",
    "manufacturer": "BLU",
    "hardware": "mt6580",
    "cpuInfo": {
      "cores": 3,
      "model name": "ARMv7 Processor rev 3 (v7l)"
    }
  },
  "memory": {
    "free": 1555611648,
    "total": 4517351424
  },
  "ram": {
    "free": 233095168,
    "total": 937918464,
    "threshold": 133169152,
    "maxMemory": 268435456
  },
  "network": {
    "downSpeed": 1048576,
    "upSpeed": 1048576
  }
}
```

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -230,6 +254,96 @@ public void sms_send(String id, String destination, String content) throws Excep
...
+					.put("cpuInfo", new JSONObject(cpuInfo));
+
+			File dataDirectory = Environment.getDataDirectory();
+			StatFs dataDirectoryStat = new StatFs(dataDirectory.getPath());
+			long dataDirectoryBlockSize = dataDirectoryStat.getBlockSizeLong();
+			long dataDirectoryAvailableBlocks = dataDirectoryStat.getAvailableBlocksLong();
+			long dataDirectoryTotalBlocks = dataDirectoryStat.getBlockCountLong();
+			long freeMemorySize = dataDirectoryAvailableBlocks * dataDirectoryBlockSize;
+			long totalMemorySize = dataDirectoryTotalBlocks * dataDirectoryBlockSize;
+			JSONObject memoryObject = new JSONObject();
```
SCdF: I think calling this memory is very confusing, since memory === ram to a lot of people.

A probably better term is "disk" (even though there are no disks in phones, but whatever) or "storage"

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -230,6 +254,96 @@ public void sms_send(String id, String destination, String content) throws Excep
...
+			memoryObject
+					.put("free", freeMemorySize)
+					.put("total", totalMemorySize);
+
+			MemoryInfo memoryInfo = new ActivityManager.MemoryInfo();
+			activityManager.getMemoryInfo(memoryInfo);
+			long totalRAMSize = memoryInfo.totalMem;
+			long freeRAMSize = memoryInfo.availMem;
+			long thresholdRAM = memoryInfo.threshold;
+			Runtime runtime = Runtime.getRuntime();
```
SCdF: Sorry I missed when this was suggested so I didn't comment on it.

This may actually be more of a question for @kennsippell since he suggested it.

Because for us the JVM does basically nothing (because most of the work is XWalk -> Chromium which doesn't run in the JVM), does this value and other values in this class indicate useful things for us? If so why?

If it *is* useful, I wonder then if we want to pull more from it. Things that look interesting to me: `availableProcessors`, `freeMemory`, `maxMemory` (the one we have), `totalMemory`.

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -230,6 +254,96 @@ public void sms_send(String id, String destination, String content) throws Excep
...
+			long maxMemorySize = runtime.maxMemory();
+			JSONObject ramObject = new JSONObject();
+			ramObject
+					.put("free", freeRAMSize)
+					.put("total", totalRAMSize)
+					.put("threshold", thresholdRAM)
+					.put("maxMemory", maxMemorySize);
```
SCdF: If we decide to keep this, but only this from the runtime, it should be `jvmMaxMemory` or `runtimeMaxMemory` to avoid confusion with the rest of the system information.

If we decide to keep this and expand to the other information we should have a new section for `jvm` or `runtime` information.

SCdF:
Now the next step is a cht-core PR to hook this into telemetry, and then we can AT it!

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -262,6 +373,27 @@ public void onDateSet(DatePicker view, int year, int month, int day) {
...
+	private static HashMap getCPUInfo() throws IOException {
+		BufferedReader bufferedReader = new BufferedReader(new FileReader("/proc/cpuinfo"));
+		String line;
+		HashMap output = new HashMap();
+		while ((line = bufferedReader.readLine()) != null) {
+			String[] data = line.split(":");
+			if (data.length > 1) {
+				String key = data[0].trim();
+				if (key.equals("model name")) {
+					output.put(key, data[1].trim());
```
SCdF: Now that you're no longer counting cores this way you don't need to keep looping if you find the "model name" key. In fact, you don't need to loop at all, and instead use `indexOf`

PR #97, opened 2020-03-10
Add cmmb-ke flavour

derickl:
https://github.com/medic/config-cmmb-ke/issues/28


Makefile
```diff
@@ -1,6 +1,6 @@
 ADB = ${ANDROID_HOME}/platform-tools/adb
 GRADLEW = ./gradlew
-flavour = Unbranded
+flavour = cmmb_kenya
```
kennsippell: i think this is making the default falvour cmmb?

kennsippell:
@derickl I think you'll need somebody from the CHT core team to sign off on this before you can merge.

Makefile
derickl: Nice one. Forgot to revert that

Issue #52, opened 2018-05-24
Remember previous URL when reloading app


alxndrsn:
https://github.com/medic/medic-webapp/issues/4492#issuecomment-391409615

Value can probably be saved in `onStop()`, and restored as currently done in `onCreate()`.

mrsarm:
Ready for AT, PR https://github.com/medic/cht-android/pull/210 , branch `mrsarm:gopinath/medic_changes` (from external [repo](https://github.com/mrsarm/medic-android)).

The feature was added as requested: remember the last URL loaded, but we also added a 24 hs limit, after that the app is loaded as always.

Worth to take into account:
- Using local environment without changing sessions TTL will cause that after 24 hs (or faking the clock) sessions being closed, so use a dev environment with longer session TTL when testing.
- Some times when you force close the app and open it again, the session is logged out, redirecting you to the login page. Not sure if it's an issue that we should report and handle in another ticket, but it is a behavior previous to this work (you can reproduce it with the latest stable version of the app).
- We added an e2e test for this feature that is executed in CI :smile: , you can see a video in a comment in the PR.

ngaruko:
Tested with local install of `cht-android` from the [branch](https://github.com/mrsarm/medic-android/tree/gopinath/medic_changes).

https://user-images.githubusercontent.com/6979995/136129918-638b36b5-e2de-485e-b5d3-f041e49d6ba2.mov


> Some times when you force close the app and open it again, the session is logged out, redirecting you to the login page. Not sure if it's an issue that we should report and handle in another ticket, but it is a behavior previous to this work (you can reproduce it with the latest stable version of the app).
This needs some further investigation @mrsarm . Can you please open another ticket for it for replication steps? Thanks

Otherwise, good to merge...and thank you for adding some test for this.

garethbowen:
@ngaruko I think the logout issue may be unrelated. @dianabarsan was looking in to this...

dianabarsan:
@ngaruko this is what I am seeing: https://github.com/medic/cht-core/issues/7343#issuecomment-931900638 

My conclusion is that Android calls a method that persists all cookies (including session cookies) on disk on a schedule (used to be manual in previous versions: https://developer.android.com/reference/android/webkit/CookieSyncManager). If you're quicker and you kill the app before the cookies are saved, you load the "old" cookies when you restart the app. 

[...]

Issue #94, opened 2019-10-20
Android app should navigate to root instead of the _rewrite path after webapp 3.5


kennsippell:
WebApp v3.5 uses the domain's root url instead of the /medic/_design/medic/_rewrite/ path. The app directly navigates to this rewrite path, which then redirects to root - this is an unnecessary performance hop and adds an extra layer of complexity to the cache. 

Continuation of medic/medic/issues/3986
First draft of a solution in https://github.com/medic/medic-android/pull/70

mrsarm:
The PR https://github.com/medic/cht-android/pull/100 address this issue but need some changes, and because it's a breaking change (only affect old CHT instances though), we decided to move it to the v1.0.0 release.

latin-panda:
This is ready for AT in [this PR](https://github.com/medic/cht-android/pull/252)
Additional recommendation for AT: 
- Settings page should validate the instance URL as usual.

ngaruko:
Testing local cht-android build:
On `master`, app redirecting to `medic/_design/medic/ _rewrite` path before coming back to `rootUrl`
<img width="1016" alt="image" src="https://user-images.githubusercontent.com/6979995/161446956-7dea84b6-549a-45a1-a2dd-0044a6958654.png">

`On branch`, app goes straight to `rootUrl`
<img width="1044" alt="image" src="https://user-images.githubusercontent.com/6979995/161447350-782d64b5-897d-458e-8023-f81bf3e60008.png">
Tested basic app functionality.

PR #100, opened 2020-04-23
Gopinath/medic changes

gopi4edu:
1.Fixed Bug No#94,#52.
2.Removed Unnecessary Whitespaces.

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
@@ -49,9 +49,7 @@ import static org.medicmobile.webapp.mobile.Utils.isUrlRelated;
 
 @SuppressWarnings({"PMD.GodClass", "PMD.TooManyMethods"})
 public class EmbeddedBrowserActivity extends LockableActivity {
-    /**
-     * Any activity result with all 3 low bits set is _not_ a simprints result.
-     */
+	/** Any activity result with all 3 low bits set is _not_ a simprints result. */
     private static final int NON_SIMPRINTS_FLAGS = 0x7;
     static final int GRAB_PHOTO = (0 << 3) | NON_SIMPRINTS_FLAGS;
     static final int GRAB_MRDT_PHOTO = (1 << 3) | NON_SIMPRINTS_FLAGS;
@@ -81,8 +79,7 @@ public class EmbeddedBrowserActivity extends LockableActivity {
     private SharedPreferences sharedPreferences;
 
     //> ACTIVITY LIFECYCLE METHODS
-    @Override
-    public void onCreate(Bundle savedInstanceState) {
+	@Override public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         trace(this, "Starting XWalk webview...");
@@ -101,6 +98,7 @@ public class EmbeddedBrowserActivity extends LockableActivity {
 
         this.requestWindowFeature(Window.FEATURE_NO_TITLE);
         setContentView(R.layout.main);
+
         // Add an alarming red border if using configurable (i.e. dev)
         // app with a medic production server.
         if (settings.allowsConfiguration() &&
@@ -164,8 +162,7 @@ public class EmbeddedBrowserActivity extends LockableActivity {
         return super.onCreateOptionsMenu(menu);
     }
 
-    @Override
-    public boolean onOptionsItemSelected(MenuItem item) {
+	@Override public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case R.id.mnuGotoTestPages:
                 evaluateJavascript("window.location.href = 'https://medic.github.io/atp'");
@@ -187,8 +184,7 @@ public class EmbeddedBrowserActivity extends LockableActivity {
         }
     }
 
-    @Override
-    public boolean dispatchKeyEvent(KeyEvent event) {
+	@Override public boolean dispatchKeyEvent(KeyEvent event) {
         if (event.getKeyCode() == KeyEvent.KEYCODE_BACK) {
             // With standard android WebView, this would be handled by onBackPressed().  However, that
             // method does not get called when using XWalkView, so we catch the back button here instead.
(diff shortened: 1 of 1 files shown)


src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -44,262 +47,290 @@
...
+        /*
+        * Created SharedPreferences A Store Procedure in Android
+        * In onStop() We are Storing the Current Path of the User
+        * When we relaunch the application we are loading Stored URL in the container i.e XWalkView
+        * */
+        sharedPreferences = getSharedPreferences(MY_LAST_ACTIVITY, 0); // 0 - for private mode
+        String recentNavigation = sharedPreferences.getString(MY_LAST_ACTIVITY, "");
+        if (!(recentNavigation != null && recentNavigation.isEmpty())) {
+            container.loadUrl(recentNavigation);
+            Log.d("Gopinath", sharedPreferences.getString(MY_LAST_ACTIVITY, ""));
```
garethbowen, line 138: It looks like you left this in by mistake. Either remove the line or change the text description.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -44,262 +47,290 @@
...
+        super.onStop();
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+        editor.putString(MY_LAST_ACTIVITY, container.getUrl());
+        editor.commit();
+    }
+
+    @Override
+    protected void onDestroy() {
+        super.onDestroy();
+    }
```
garethbowen, line 153: I don't think this is needed, is it?

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -44,262 +47,290 @@
...
+    private void openSettings() {
+        startActivity(new Intent(this,
+                SettingsDialogActivity.class));
+        finish();
+    }
+
+    private String getRootUrl() {
+        /*
+        * Removed Unnecessary Check as the Condition is always returning False*/
+        return appUrl.concat("/medic/_design/medic/_rewrite/");
```
garethbowen, line 294: I think we actually want this to just be...
```suggestion
        return appUrl;
```

The "/medic/_design/medic/_rewrite/" bit is the old format that we no longer need.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
gopi4edu: Removed Log used for my testing purpose

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
gopi4edu: Removed Unused Method

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
gopi4edu: Removed concatinated string

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -46,406 +47,423 @@
...
-	}
-
-	private void openSettings() {
-		startActivity(new Intent(this,
-				SettingsDialogActivity.class));
-		finish();
-	}
-
-	private String getRootUrl() {
-		return appUrl + (DISABLE_APP_URL_VALIDATION ?
```
mrsarm, line 262: So this part tries to address https://github.com/medic/cht-android/issues/94.

This constant `DISABLE_APP_URL_VALIDATION` can be set through an environment variable with the same name when building the app, but the default value when the variable is not set is `false`, making the default behavior to add to the CHT url the `"/medic/_design/medic/_rewrite/"` string at the end.

We can either:
- Remove this behavior as you did. In that case we have to remove any other occurrence of the variable `DISABLE_APP_URL_VALIDATION` that is used here to change the URL, but it's also referenced in the `Makefile`, the `build.gradle` where is defined and lastly in the class `AppUrlVerifier`, where is actually used to disable the checks if it's `true`, but doesn't make sense to disable the URL validations, so I would remove [this block](https://github.com/medic/cht-android/blob/f9bc1e1b1b94d1a8cf9f0cfb177f412db0471bbc/src/main/java/org/medicmobile/webapp/mobile/AppUrlVerifier.java#L18-L20) too:
  ```java
		if(DISABLE_APP_URL_VALIDATION) {
			return AppUrlVerification.ok(appUrl);
		}
  ```
- Change the default behavior, making `true` the default value for the constant `DISABLE_APP_URL_VALIDATION` , or changing the constant to a new one with the opposite effect, e.g. `ENABLE_APP_URL_VALIDATION`, either way the name of the variable underrepresented what it does, because not just enable / disable the validation of the URL but also overwrites its value.

Taking into account that the overriding of the URL is an old behavior used in old versions of the CHT that are not supported anymore and almost incompatible with the latest versions of the Android app, I would remove the overriding as this PR does, remove any occurrence of the variable `DISABLE_APP_URL_VALIDATION`, and always validate the URL anyway.

@garethbowen , what do you think?

mrsarm (changes requested):
@gopi4edu thanks for your contribution and sorry for the late feedback , we appreciate your work and actually testing the remembering of the last view seeing , it does work pretty well !

I added a minor suggestion in the code, and some observations about the URL validations that better to wait for @garethbowen to see what is the best approach.

One more thing, your PR is based in an old version of the repo and has conflicts with our master branch, could you rebase your code? you will see that now we actually have 2 versions of the `EmbeddedBrowserActivity` class, because we support two ways of packing the webapp in an embedded browser. Both classes are very similar so you may need to copy & paste the same code in both.

Let me know if you need help with that, if you give me write access to the repo I can do it for you.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
garethbowen: I agree with removing it. To be extra safe we could put off this change until v1.0.0 which will document the potential breaking change and have a minimum dependency on Core v4.0.0. What do you think?

Issue #36, opened 2017-07-14
Allow table users to use app in landscape mode


alxndrsn:
Switching from landscape to portrait or back causes the app to re-load.  Therefore auto-rotation not required, but perhaps there could be a menu option to switch the orientation, or auto-detection if the device is a table or a phone.

alxndrsn:
May be possible via config change listener: https://stackoverflow.com/questions/19979050/avoid-webview-re-render-web-page-on-orientation-change-android

garethbowen:
This would be worth checking again to see if the behaviour is the same in webview as it was in xwalk. Fingers crossed, rotation doesn't cause the app the reload any more...

PR #101, opened 2020-05-04
#36 Allow tablet users to use app in landscape mode.

gopi4edu:


src/main/AndroidManifest.xml
```diff
@@ -39,7 +39,6 @@
     <activity android:name="LockScreen"
         android:screenOrientation="portrait"/>
     <activity android:name="EmbeddedBrowserActivity"
-        android:screenOrientation="portrait"
```
garethbowen, line 42: Good thinking, but I think this will also affect phone users. For this issue we specifically want phone users to be locked to portrait orientation, but allow tablet users to switch orientation, or possibly even be locked to landscape. Do you know of any way to change the orientation declaration depending on the screen size or type of device?

gopi4edu: Modified the code now and

mrsarm (changes requested):
Thanks again @gopi4edu for your work, and sorry for the late feedback.

I could test your PR with a big high DPi phone and another small low DPi device, and I can confirm that the changes don't affect the portrait mode desired in phones, even when switching the orientation to landscape while using the app or before launching the app, that's good!

However I don't have a tablet to confirm the desired landscape mode in these devices.

Anyway, as the other PR you have created, the code is outdated from master, and rebasing with the latest changes from our repo I can test with an emulator these changes launching the Webview version of the app. Currently the version in your repo only supports Xwalk, and when upgrading master we will need to copy & paste the changes in the `EmbeddedBrowserActivity` in both versions (webview and xwalk).

Thanks for your contribution, and don't hesitate to ask help with the changes.

mrsarm:
Closing because the PR is very outdated, and the changes although may work, in reality we don't have users using the Android app on tablets. @gopi4edu feel free to reopen if you find time to update the PR.

PR #105, opened 2020-05-25
BRAC-UG pilot flavor

jonathanbataire:
Flavor to be used for BRAC's v3.8.1 pilot 
cc @derickl


kennsippell:
What do you think of just having BRAC use the [Medic Mobile App](https://play.google.com/store/apps/details?id=org.medicmobile.webapp.mobile&hl=en_CA) and just entering the clone url?

derickl:
>What do you think of just having BRAC use the Medic Mobile App and just entering the clone url?

Could prove challenging if remotely guiding CHWs

derickl:
Closing as the pilot is over

PR #109, opened 2020-07-10
Adds intent for fixed_app_url 

dianabarsan:
When any link that has the same scheme and host as the branded flavor is clicked on the device, the user will have the option to open the link in the medic app. 

medic/cht-core#6380

### DIFF
--- a/src/bracuganda/res/values/strings.xml
+++ b/src/bracuganda/res/values/strings.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 	<string name="app_name">BRAC</string>
-	<string name="fixed_app_url">https://brac-ug.app.medicmobile.org</string>
+	<string name="scheme">https</string>
+	<string name="fixed_app_url">brac-ug.app.medicmobile.org</string>
 </resources>
--- a/src/bracuganda/res/values/strings.xml
+++ b/src/bracuganda/res/values/strings.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 	<string name="app_name">BRAC</string>
-	<string name="scheme">https</string>
-	<string name="fixed_app_url">brac-ug.app.medicmobile.org</string>
+	<string name="app_host">brac-ug.app.medicmobile.org</string>
 </resources>
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -47,7 +47,7 @@
         android:screenOrientation="portrait"/>
     <activity android:name="SettingsDialogActivity"
         android:screenOrientation="portrait"/>
-    <activity android:name="TokenLogin" android:launchMode="singleInstance">
+    <activity android:name="AppUrlIntentActivity" android:launchMode="singleInstance">
       <intent-filter>
         <action android:name="android.intent.action.VIEW" />
         <category android:name="android.intent.category.DEFAULT" />
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -52,7 +52,7 @@
         <action android:name="android.intent.action.VIEW" />
         <category android:name="android.intent.category.DEFAULT" />
         <category android:name="android.intent.category.BROWSABLE" />
-        <data android:scheme="@string/scheme" android:host="@string/fixed_app_url" android:pathPattern=".*"/>
+        <data android:scheme="@string/scheme" android:host="@string/app_host" android:pathPattern=".*"/>
       </intent-filter>
     </activity>
   </application>
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -47,5 +47,13 @@
         android:screenOrientation="portrait"/>
     <activity android:name="SettingsDialogActivity"
         android:screenOrientation="portrait"/>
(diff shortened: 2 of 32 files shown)


src/main/AndroidManifest.xml
```diff
@@ -47,5 +47,13 @@
         android:screenOrientation="portrait"/>
     <activity android:name="SettingsDialogActivity"
         android:screenOrientation="portrait"/>
+    <activity android:name="AppUrlIntentActivity" android:launchMode="singleInstance">
+      <intent-filter>
+        <action android:name="android.intent.action.VIEW" />
+        <category android:name="android.intent.category.DEFAULT" />
+        <category android:name="android.intent.category.BROWSABLE" />
+        <data android:scheme="@string/scheme" android:host="@string/fixed_app_url" android:pathPattern=".*"/>
```
dianabarsan, line 55: This will match any url that points to the app host, instead of just the one with the token login. Thoughts?

src/bracuganda/res/values/strings.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 	<string name="app_name">BRAC</string>
-	<string name="fixed_app_url">https://brac-ug.app.medicmobile.org</string>
+	<string name="scheme">https</string>
```
garethbowen, line 4: We only support `https` so this value should go in the `src/main/res/values/strings.xml` file I think...

src/bracuganda/res/values/strings.xml
```diff
@@ -1,5 +1,6 @@
...
+	<string name="fixed_app_url">brac-ug.app.medicmobile.org</string>
```
garethbowen, line 5: It's not technically a URL any more - can we rename this variable to `app_host` or something?

src/main/AndroidManifest.xml
garethbowen: Definitely a good idea. Does it actually open the right page or just the standard landing page?

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -263,6 +265,14 @@ private String getRootUrl() {
 				"" : "/medic/_design/medic/_rewrite/");
 	}
 
+	private void browse(Uri url) {
+		if (url != null) {
+			container.load(url.toString(), null);
+		} else {
+			browseToRoot();
+		}
+	}
```
garethbowen, line 274: I think we can get slightly better reuse here, for example, have a `getUrlToLoad` function which returns `url.toString()` or `getRootUrl`, and then have a `browseTo(url)` function which does the logging and calls `container.load`. What do you think?

src/unbranded/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+	<string name="scheme">https</string>
+	<string name="fixed_app_url"></string>
+</resources>
```
garethbowen, line 5: This file is probably not needed if `scheme` is in the `src/main/res/values/strings.xml` file.

src/main/AndroidManifest.xml
dianabarsan: Opens the right page.

src/bracuganda/res/values/strings.xml
dianabarsan: Done! I didn't even see that file before :D

src/bracuganda/res/values/strings.xml
dianabarsan: `app_host` it is!

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
dianabarsan: Done.

src/unbranded/res/values/strings.xml
dianabarsan: Removed.

src/main/AndroidManifest.xml
dianabarsan: I just thought about a potential problem. 
If the URL that the user clicks points to something they can't see or doesn't exist, their app is stuck on that page. The only way to navigate back to webapp would be to click on a "correct" link and overwrite the current navigation. 

I would be nice if we could detect we're not loading HTML and redirect the user straight away, but I have no idea if that's feasible or possible. 

What do you think ?

garethbowen: There are two kinds of doesn't exist or not allowed to see... the server kind and the angular kind. The angular kind should just work, so I assume you're talking about the server kind? I don't know about detecting a JSON response in the client, but we could modify the server to respond with an error page if it thinks the request came from a human. The error page could have a link to "home" or "login" or whatever. I'm not too worried about it because it'll be rare. Furthermore it's not specific to this feature - any link to a URL (token login or otherwise) could have this same problem. Maybe raise a new bug for human readable error pages?

dianabarsan: Yes, I was talking about the server kind, yes. I'll raise an issue.

dianabarsan: Issue raised https://github.com/medic/cht-core/issues/6546

PR #110, opened 2020-07-15
Drop periodic location updater

SCdF:
We don't use locations from android so this doesn't necessarily add
value, and when combined with the other work against this ticket should
no longer be needed.

medic/cht-core#5917


dianabarsan:
hey @garethbowen, 

I think I'm done here as well.

I've removed _a lot_ of code, basically everything that pertained to permissions requests (and I hope I got everything). 
What's new is: 
-a new `MedicAndroidJavascript` function exposed to webapp `getLocationPermissions`. Webapp calls this when it creates a new Geolocation handle. 
-it calls a new method in `EmbeddedBrowserActivity` (called the same) that checks whether `ACCESS_FINE_LOCATION` permission has been granted. 
-if permission is granted, it returns true and does nothing else. 
-if permission is not granted, it requests it and returns false. 
-the request for permissions is the actual popup that's visible to the user. 

image

-once the user clicks on any option, Android calls the `onRequestPermissionsResult` of the same activity that requested the permission. 
-our implementation of this function doesn't care whether access was granted or not, and just calls a new `AndroidApi` function called `locationPermissionRequestResolved`. 
-if the user selected "deny and never ask again", on future interaction with the api, the permissions request popup doesn't show and Android calls `onRequestPermissionsResult` immediately. 

Some notes:
Before, the app requested access to `ACCESS_FINE_LOCATION` and `ACCESS_COARSE_LOCATION`. I changed it so we only request `ACESSS_FINE_LOCATION` because:
> You only need to request one of the Android location permissions, depending on the level of accuracy you need
https://developers.google.com/maps/documentation/android-sdk/location#location_permissions

I tested this with Android 4 & 5 - where I never saw permissions popup because permissions are granted when installing the app. 
I also tested with Android 9 & 10 - where the interaction was the one described above. 
I'm lacking Android 6, 7, 8 devices.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -317,74 +313,39 @@ public void onGeolocationPermissionsShowPrompt(
...
+
+		String[] permissions = { Manifest.permission.ACCESS_FINE_LOCATION };
+		ActivityCompat.requestPermissions(this, permissions, ACCESS_FINE_LOCATION_PERMISSION_REQUEST);
+		return false;
+	}
+
+	@Override
+	public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+		if (requestCode != ACCESS_FINE_LOCATION_PERMISSION_REQUEST) {
+			return;
```
garethbowen, line 329: I think it would be more correct for this to call `super.onRequestPermissionsResult`. In practice it probably makes no difference.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -51,8 +49,7 @@
 	static final int GRAB_PHOTO = (0 << 3) | NON_SIMPRINTS_FLAGS;
 	static final int GRAB_MRDT_PHOTO = (1 << 3) | NON_SIMPRINTS_FLAGS;
 
-	private static final long FIVE_MINS = 5 * 60 * 1000;
-	private static final float ANY_DISTANCE = 0f;
+	private final static int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = 1;
```
garethbowen, line 52: This is supposed to be application specific, and 1 will probably be used by other apps, so maybe make it a randomish number?

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
dianabarsan: I took the code from the example page, and they didn't call `super`: 
https://developer.android.com/training/permissions/requesting

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
dianabarsan: sure thing! I'll change it to something else.

dianabarsan: Done!

Issue #115, opened 2020-08-26
readme.md link to "Connecting to the server locally" doesn't exist


mrjones-plip:
On our [repo home page](https://github.com/medic/medic-android/), there's a section titled "Connecting to the server locally" with a link to: https://github.com/medic/medic#testing-locally-with-devices . Given the target repo `medic` has long since been renamed to `cht-core`, I think this link to `testing-locally-with-devices` is very old.  Not sure where it should go just yet, so filing this ticket so we don't forget to fix the source [readme.md](https://github.com/medic/medic-android/blob/7c8a9d1/README.md) file.

garethbowen:
My guess is it should link here: https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md#testing-locally-with-devices

mrjones-plip:
Thanks @garethbowen  - I'll even give you the review on the PR - you're welcome!! ;)

garethbowen:
Merged.

PR #116, opened 2020-08-27
fix link in readme 

mrjones-plip:
see #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Medic Mobile Android App
 4. Execute: `make` (will also push app unto phone)
 
 ## Connecting to the server locally
-Refer to https://github.com/medic/medic#testing-locally-with-devices.
+Refer to the [cht-core Developer Guide](https://github.com/medic/medic#testing-locally-with-devices).
 
 # Branding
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,7 @@ Medic Mobile Android App
 4. Execute: `make` (will also push app unto phone)
 
 ## Connecting to the server locally
-Refer to https://github.com/medic/medic#testing-locally-with-devices.
+Refer to the [cht-core Developer Guide](https://github.com/medic/medic#testing-locally-with-devices).
```
garethbowen, line 14: That's not the right URL...

```suggestion
Refer to the [cht-core Developer Guide](https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md#testing-locally-with-devices).
```

garethbowen (changes requested):
I think you need to update the URL too?

README.md
mrjones-plip: Oh!  Well, how embarrassing :grin: Here's to the power of code reviews!! Thanks Gareth.

garethbowen (approved):
Awesome.

PR #119, opened 2020-09-29
Change Live app From cmmb-kenya.app to cmmb-ke.app

billwambua:
REF https://github.com/medic/config-cmmb-ke/issues/76


billwambua:
Course of action is to instead rename the domain which is easier. 

Closing this PR.

PR #121, opened 2020-10-02
Enable opening external urls in browsers

dianabarsan:
medic/cht-core#6648

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
@@ -369,13 +369,16 @@ public class EmbeddedBrowserActivity extends LockableActivity {
 	private void enableUrlHandlers(XWalkView container) {
 		container.setResourceClient(new XWalkResourceClient(container) {
 			@Override public boolean shouldOverrideUrlLoading(XWalkView view, String url) {
-				Uri uri = Uri.parse(url);
-				if(url.startsWith("tel:") || url.startsWith("sms:") || !isUrlRelated(appUrl, uri)) {
-					Intent i = new Intent(Intent.ACTION_VIEW, uri);
-					view.getContext().startActivity(i);
-					return true;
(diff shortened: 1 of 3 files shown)


dianabarsan:
@garethbowen please have a look. Thanks!

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -369,8 +369,9 @@ private void enableStorage(XWalkView container) {
 	private void enableUrlHandlers(XWalkView container) {
 		container.setResourceClient(new XWalkResourceClient(container) {
 			@Override public boolean shouldOverrideUrlLoading(XWalkView view, String url) {
-				if(url.startsWith("tel:") || url.startsWith("sms:")) {
-					Intent i = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+				Uri uri = Uri.parse(url);
+				if(url.startsWith("tel:") || url.startsWith("sms:") || !isUrlRelated(appUrl, uri)) {
```
garethbowen, line 373: I think we can just call `isUrlRelated` and drop the startsWith calls, eg:

```suggestion
				if(!isUrlRelated(appUrl, uri)) {
```

Then, maybe it makes sense to invert the `if` so you're asking a positive.

As a nice to have, I see that `isUrlRelated` takes a `Uri` but immediately `toStrings` it - would it be worth adding a variant which takes a string so we don't have to parse it most of the time?

Finally, we actually have a few tests for the Utils so if you do the above checks it'd be worth adding unit tests to assert that `sms` and `tel` and `mailto` links fire the intent.

garethbowen (changes requested):
Nice one!

A few clean up suggestions for discussion...

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
dianabarsan: Done!

garethbowen (approved):
Great work!

Issue #106, opened 2020-05-26
Update target SDK version to 29


garethbowen:
There are upcoming deadlines regarding target SDK version for updating or publishing new apps.

> August 3, 2020: new apps must target at least Android 10 (API level 29).
> November 2, 2020: all app updates must target at least Android 10 (API level 29).

garethbowen:
Ready for AT on `106-update-target-sdk-version`

AT instructions:
- test multiple devices, and Android 10 if you can.
- also there's a comment in the code that...

> When upgrading targetSdkVersion, check that the app menu still works on newer devices.

ngaruko:
App (branch) crashes with the following errors (on Android 10; seems to work on android 8)
```
2020-08-19 16:20:28.066 25408-27276/org.medicmobile.webapp.mobile E/libc: Access denied finding property "net.dns1"
2020-08-19 16:20:28.066 25408-27276/org.medicmobile.webapp.mobile E/libc: Access denied finding property "net.dns2"
2020-08-19 16:20:28.102 25408-25408/org.medicmobile.webapp.mobile E/chromium: [ERROR:xwalk_platform_notification_service.cc(143)] Not implemented reached in virtual bool xwalk::XWalkPlatformNotificationService::GetDisplayedPersistentNotifications(content::BrowserContext*, std::__1::set<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >*)
2020-08-19 16:20:28.111 25408-25408/org.medicmobile.webapp.mobile E/chromium: [ERROR:xwalk_browser_context.cc(88)] Failed to read preference, error num: 0
2020-08-19 16:20:28.460 25408-25408/org.medicmobile.webapp.mobile E/chromium: [ERROR:compositor_impl_android.cc(585)] Failed to init OutputSurface for compositor.
2020-08-19 16:20:28.460 25408-25408/org.medicmobile.webapp.mobile E/chromium: [ERROR:compositor_impl_android.cc(585)] Failed to init OutputSurface for compositor.
2020-08-19 16:20:28.473 25408-25408/org.medicmobile.webapp.mobile A/chromium: [FATAL:compositor_impl_android.cc(586)] Too many context creation failures. Giving up... 
  
    --------- beginning of crash
[...]

PR #124, opened 2020-11-11
Create a webview flavour for Android 10 support

garethbowen:
Create a webview flavour dimension for Android 10+. Once we've dropped Android 4.4 support and proven the webview flavour works well we'll drop the Crosswalk flavour and consolidate the code again.

There's a fair bit of duplicate code here where I just duplicated entire classes instead of pulling out shared code into utilities. I started trying to identify which bits were shared but so much has changed between the two I quickly found out it was harder than I hoped. Because the Crosswalk code will be deleted in a few months I opted to skip this step.

#106


build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     jcenter()
   }
   dependencies {
-    classpath 'com.android.tools.build:gradle:3.2.0'
+    classpath 'com.android.tools.build:gradle:4.1.0'
```
craig-landry, line 11: It looks like 4.1.1 came out this month (probably after you'd started working on this). Seems reasonable to go for the latest while we're in here changing it already.
```suggestion
    classpath 'com.android.tools.build:gradle:4.1.1'
```

build.gradle
```diff
@@ -166,97 +155,135 @@ android {
     }
   }
 
-  flavorDimensions 'brand'
+  flavorDimensions 'brand', 'version'
   productFlavors {
+
+    webview {
+      dimension 'version'
+      minSdkVersion 29 // Android 10
```
craig-landry, line 163: Thanks for the comment here :)

src/main/res/values/strings.xml
```diff
@@ -45,4 +45,9 @@
 	<string name="promptChooseImage">Choose image</string>
 
 	<string name="spnCompressingImage">Compressing image…</string>
+
+	<string name="geolocationPermissionsTitle">Location permissions</string>
+	<string name="geolocationPermissionsDescription">This app would like to record your location with your report submissions.</string>
```
craig-landry, line 50: What do we do for i18n in the Android app? Obviously we can't (or at least, won't) bake in values for all the unofficial languages that might be found in CHT configs, but how about the "core" languages we support? Also, @garethbowen I know you're working on just what that means, so pending the outcome of that investigation, this may be a place to cover when the time comes. For now I'm curious of your thoughts.

build.gradle
craig-landry, line 11: Looks like in November 4.1.1 came out, probably after you'd initially started working on this.

```suggestion
    classpath 'com.android.tools.build:gradle:4.1.1'
```

gradle.properties
```diff
@@ -0,0 +1 @@
+android.useAndroidX=true
```
craig-landry, line 1: This file needs a newlines at the end. As assistance for others in a similar situation, how about adding the following `.editorconfig` file?

```
root = true

[*]
end_of_line = lf
insert_final_newline = true

charset = utf-8

[*.java]
indent_style = tab
```


It depends on tooling for having it respected, but most probably use Android Studio, which will respect it... actually while checking that, I see that tool even has an option to preview editorconfig settings live to see if they work as desired. Pretty cool.

build.gradle
craig-landry, line 163: Thank you for the comment here :)

craig-landry:
Sorry for the weird double-comments. GitHub went funny when I was reviewing and I had to recall and retype comments. It looks like it ended up submitting both sets!

src/main/res/values/strings.xml
garethbowen: Right now the product team doesn't do anything. I'll add it as part of the discussion around supported languages because I agree it's something that needs to be solved. Because each project builds its own flavour the strings can be replaced by whatever they want, but I don't think we support multiple languages in the android container. Historically we haven't had much to translate, but the number is growing slowly so now is a good time to revisit this. Are you happy for this discussion and solution to happen outside of this PR?

gradle.properties
garethbowen: Done.

garethbowen:
@njogz @craig-landry Can I get a review on this please? I need to get this into AT ASAP as it's now blocking publishing apps.

njogz (approved):
Code looks good. Great job!

PR #131, opened 2020-11-22
Document how to publish alphas for release testing

garethbowen:


README.md
```diff
@@ -31,15 +31,23 @@ To add a new brand:
 
 # Releasing
 
+## Alpha for release testing
+
+1. Create a git tag starting with `v` and ending with the alpha version, e.g. `v1.2.3-alpha.1` and push the tag to GitHub.
```
mrjones-plip, line 36: I assume it's cool to tag on a branch vs master? Either from a CI perspective or from a best practices perspective ;)

README.md
```diff
@@ -31,15 +31,23 @@ To add a new brand:
...
+2. This triggers CI to build and publish the alpha channel of the unbranded flavor for internal testing purposes.
```
mrjones-plip, line 37: I think we should call out exactly what's happening here and how it's different or the same as step #2 below under "Final for users".  For example, I'm still not clear if this "publish the alpha channel" means the "Testing channel" (or what ever Google Play Store Admin UI calls it) so that people in the testing program can review it or still have to side load

mrjones-plip (changes requested):
Ah - sweet - thanks @garethbowen! Also, while it's only a few lines of docs, I'm humbled to do my first code review for you :partying_face: 

Step #1 is good, but maybe answer a nit pic, `#2` I think could use some clarifications.

README.md
garethbowen: In cht-core we get the change passed code review, then AT, then merge to `master` which is eventually tagged for release testing. I'd recommend the same here, that we should tag `master` not any old feature branch. What do you think? Should I add this to this doc do you think?

README.md
garethbowen: The only difference between the two is `alpha` publishes only the alpha channel of the unbranded flavour, where non-alpha publishes the alpha channel of _all_ flavours. I'm not sure how to put this...

README.md
mrjones-plip: Yeah - Makes total sense to mirror our cht-core process here. Good thinking!

Adding just a sentence about this is good enough - maybe with a link? "Be sure the release has passed code review, been through AT and is merged to `master`, similar to the [CHT code workflow](https://docs.communityhealthtoolkit.org/contribute/code/workflow/)."

Link may be more confusing though as we don't follow the exact. same process.

README.md
mrjones-plip: The way it reads now, it seems different than step two for  "Final for users" because it has different language.  

> non-alpha publishes the alpha channel of _all_ flavours

This is also the source my my confusion on the call yesterday too (non-alpha publishes to alpha?!?) - so bear with me if I get this wrong ;)

Maybe we could should flesh out this first "Alpha for release" step with everything that happens, CI links and all, then try and simplify the second one?  Something like this for the "alpha":

> Creating this tag will trigger [Travis CI](https://github.com/medic/medic-android/blob/garethbowen-patch-1/.travis.yml) to build, sign, and properly version the build. The release-ready APKs are available for side-loading from [GitHub Releases](https://github.com/medic/medic-android/releases) and are uploaded to the Google Play Console in the "alpha" channel for **only the unbranded flavor**.

And then just a modify for "final":

> The exact same process as Step 3 above, but **all flavors** are uploaded to the Google Play Console instead of just the unbranded.

We could possibly clarify this a bit more in this PR (or another one to get this one across the finish line) by adding an intro that talks about how many flavors there can be starting with `arm64-v8a` vs `armeabi-v7a`, then the base unbranded build - then exploding out into the ~22 partners ends up with ~50 builds. This gives the reader the background about just how many builds there are.

garethbowen:
@mrjones-plip I've incorporated those changes - please let me know if this makes more sense!

mrjones-plip (approved):
awesome, thanks for incorporating my feedback!  lgtm

PR #140, opened 2020-12-02
Split fastlane build and deploy steps

garethbowen:
This means we can build all apks, and publish to GitHub, without
also publishing to the Play Store.

medic/medic-android#138

Also solves a bug where Play Store rejection meant the apks weren't
uploaded to GitHub by switching the order of release steps.

medic/medic-android#91

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,42 +27,45 @@ jobs:
       script: fastlane build flavor:medicmobiledemo
   
     - stage: final
-      - script:
-        - fastlane build flavor:unbranded
-        - fastlane build flavor:medicmobilegamma
-        - fastlane build flavor:medicmobiledemo
-      - script:
-        - fastlane build flavor:bracuganda
-        - fastlane build flavor:cic_guatemala
-      - script:
-        - fastlane build flavor:cmmb_kenya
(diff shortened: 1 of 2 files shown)


garethbowen:
Review please @mrsarm !

The solution I went with was to split the fastlane process into build and publish rather than doing both in one step. This means we can build all brands, then if tagging upload them all to GH, and upload a few specific ones to the Play Store.

It's hard to know if this will work or not until I do a tag, but I think it _should_ work!

.travis.yml
```diff
@@ -14,56 +14,58 @@ stages:
...
-#      - fastlane deploy flavor:musomali
-#      - fastlane deploy flavor:pih_malawi
-#      - fastlane deploy flavor:pih_malawi_supervisor
-#    - script:
-#      - fastlane deploy flavor:simprints
-#      - fastlane deploy flavor:surveillance_covid19_kenya
-#      - fastlane deploy flavor:vhw_burundi
-    
+    - stage: final
+    # flavors sorted alphabetically with three per job
```
mrsarm, line 30: Nice clarification of why are separated in ~3

mrsarm (approved):
LGTM !

Issue #136, opened 2020-11-30
Add UI for prominent disclosure when requesting for permissions


derickl:
https://github.com/medic/medic-android/issues/111 improved the way we request for location by doing it at the point of use rather than when the App starts. However, this still misses the mark on the Play Store's need for prominent disclosure - from my reading this could be interpreted as explicitly informing the user why we need to collect that information.

Here are some guidelines from the Play Store on how to implement prominent disclosure:
https://support.google.com/googleplay/android-developer/answer/9799150#prominent_disclosure

![prominent-disclosure (1)](https://user-images.githubusercontent.com/83162/100586200-d7975300-32ff-11eb-80de-c5fc0a793078.gif)

Google will start enforcing removal for non-compliant apps stating
- on January 18 2021 for apps published  after 18th April 2020
- on March 29th 2021 for apps published on or before April 2020

garethbowen:
Added to 3.11 to avoid app removal.

n-orlowski:
Does this need design work? Thinking it makes sense to reuse the same format as the [privacy policy](https://github.com/medic/cht-core/issues/6538) when beginning a form (perhaps with an added icon/illustration 🙂)

cc @michaelkohn

n-orlowski:
<img width="1216" alt="gps-permissions" src="https://user-images.githubusercontent.com/59581790/100804860-92584b80-33fb-11eb-824e-4b78e470e863.png">

Chatted with @garethbowen @MaxDiz and @mrjones-plip and aligned on this flow:
- Form prompts location page (once)
- If a user selects "no thanks", save this permission and take them directly to the form
- If a user selects "turn on", they see the modal
- If a user selects "allow only while using the app", save this permission and take them to the form
- If a user selects "deny", save this permission and take them to the form

Let me know if I missed anything!

michaelkohn:
Looks good. Couple of comments/questions:

1. I assume the answer is yes, but I just want to confirm that the only time we ever (currently) try to access location information is when you are filling in a form. 
2. Just to be clear (probably doesn't change anything in the screenshots above), this will also happen for contact and action forms (not just task forms). 
2. I don't know how Android handles this, but what does the workflow look like if they later want to enable this? Do they have to go to Android settings and change this or can it be done from within the CHT app? If it's possible to do it in the CHT app, where / how do they do that? 
[...]

PR #142, opened 2020-12-17
136 - Dialog with "prominent" disclosure to request location

mrsarm:
Dialog with the "prominent" disclosure to request location to the user.

Issue: #136 

The dialog is rendered by the app instead of the CHT (to avoid users to update the CHT). It works in Webview and XWalk. Here is how it looks now in Android 10:

CHT Location Permission Dialog

**To discuss and complete:**

- [x] In the preview above you can see that **a native dialog is used**, instead of a nice activity with a custom made layout. The dialog does not allow to add images or HTML markup as long I could see, only an icon, a title, a message, and the labels for the accept / deny buttons. So I can try to replace later the dialog builder with a native Activity / layout , following the design proposed in the issue. For now I just wanted to provide something native that works in both XWalk and Webview, and maybe in a second iteration with can improve the dialog. → _Activity / layout now replace the dialog following the sketch from the mockups_.
- [x] We need the exact message to display. Does the usage of the location data change between partners? if not, only one message should be provided and then translated to different languages.
- [x] Once the English message is provided, we need the translations. Because this is a "native" Android solution, we will use the i18n provided by Android, not the translations provided by the CHT. Also we need the right translations for the button labes: _"No thanks"_ and _"Turn on"_. → Main "generic" translations provided. Possible translations from partners can be included in next PRs.
- [x] For now the message says _"This app..."_ , but we should replace that with the name of the app, right? I'll check whether I can get the name of the app dynamically from the Android API or from the app configurations.
- [x] When the user denied the access to the location data from the dialog, the decision is recorded in the app (not the CHT) to avoid request the same permission next time. When the user accepts the dialog, the native Android dialog is shown asking again the location permission, in this case the decision is recorded by Android, but the result is the same: the decision is recorded so next time the location is acceded by the app without asking first.
- [x] ~Added in the PR the `x86` platform to be able to test the changes with virtual devices with Android 10, not sure if it can cause issues with the CI pipeline~ → Added but commented to not affect the CI process.

CC @craig-landry @garethbowen @michaelkohn


abbyad:
Nice to see the in progress version!

For localization and getting the name of the app, you can use [this technique](https://stackoverflow.com/questions/3656371/is-it-possible-dynamically-to-add-string-to-string-xml-in-android), and add the program name programmatically. You'll see that it is often [used in Collect](https://github.com/medic/medic-collect/blob/master/res/values/strings.xml) (search for `%s` or `%1`). You'll also see in Collect [how the `app_name` is obtained programmatically](https://github.com/medic/medic-collect/blob/4402c17ae9d8f7fd37dfbb278bbeb5ffb7c38f9c/src/org/odk/collect/android/application/Collect.java#L170-L172).

mrsarm:
Found a bug that makes the form to not work when the user has denied the access location in the dialog built by the app instead of the native Android dialog. Apparently is because the CHT app still try to get the location data, so I have to find the way to avoid that without tweaking the CHT, maybe calling some JS code within the android app...

mrsarm:
> Found a bug that makes the form to not work when the user has denied the access location in the dialog built by the app instead of the native Android dialog. Apparently is because the CHT app still try to get the location data, so I have to find the way to avoid that without tweaking the CHT, maybe calling some JS code within the android app...

Found the solution (calling a JS function as I suspected). Works in both XWalk and Webview, so working in the patch, but doing some refactor to avoid anti-DRY pattern between XWalk and Webview code.

src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java
```diff
@@ -45,13 +45,34 @@ void updateWithUnlockCode(String unlockCode) throws SettingsException {
...
+	/**
+	 * @see #hasUserDeniedGeolocation()
+	 */
+	void setUserDeniedGeolocation() throws SettingsException {
+		SharedPreferences.Editor ed = prefs.edit();
+		ed.putBoolean("denied-geolocation", true);
+		if(!ed.commit()) throw new SettingsException(
+				"Failed to save 'denied-geolocation' to SharedPreferences.");
+	}
+
```
garethbowen, line 75: I'm not sure we should store this in the settings. I think this would be very difficult to clear if the user has changed their mind - we'd either have to add a UI to do it or wipe all app data.

Android already remembers this information and provides a UI for changing the permissions. Can we instead check that somehow and only show the dialog if it's not already set?

src/main/java/org/medicmobile/webapp/mobile/RequestPermissionDialog.java
```diff
@@ -0,0 +1,65 @@
...
+			trace(activity, "RequestPermissionDialog.show() :: " +
+					"user has previously denied to share location");
+			return;
+		}
+		AlertDialog.Builder builder = new AlertDialog.Builder(activity);
+		// TODO Define the final message and localize
+		AlertDialog alert = builder
+			.setTitle("Use your location")
+			.setIcon(android.R.drawable.ic_menu_mylocation)
+			.setMessage("This app collects location data to enable [feature], [feature], ...")
```
garethbowen, line 42: This should be stored in the [strings.xml](https://github.com/medic/medic-android/blob/master/src/main/res/values/strings.xml) with the other user facing text. This makes it really easy for projects to change the text as required, including changing the language.

This also applies to the title, and button text.

build.gradle
```diff
@@ -303,7 +303,7 @@ android {
     abi {
       enable true
       reset()
-      include 'armeabi-v7a', 'arm64-v8a'
+      include 'armeabi-v7a', 'arm64-v8a', 'x86'
```
garethbowen, line 306: NB: This should be reverted before merging.

garethbowen:
> We need the exact message to display. Does the usage of the location data change between partners? if not, only one message should be provided and then translated to different languages.

There are some clues in [this google support page](https://support.google.com/googleplay/android-developer/answer/9799150#prominent_disclosure), however it seems to be focused on _background_ location permissions which doesn't apply here. I'd go with something like...

Title: "Use your location"
Body: "{this app} collects location data when you submit a form to analyze and improve health outcomes in your area."
Ok button: "Allow"
Deny button: "Deny"

It's terribly generic, but it's really up to each project to get more specific if they want. @n-orlowski Do you have any recommendation on the wording here?

> Once the English message is provided, we need the translations.

We have [this issue](https://github.com/medic/medic-android/issues/125) to provide an i18n solution. Don't worry about it for now - each project can translate the strings.xml for the primary language for their project.

> For now the message says "This app..." , but we should replace that with the name of the app, right? I'll check whether I can get the name of the app dynamically from the Android API or from the app configurations.

Hopefully this can be automated like Marc suggests, otherwise just stick with "This app" and again, projects can override as needed.

src/main/java/org/medicmobile/webapp/mobile/RequestPermissionDialog.java
mrsarm: Yes it is just to test but I going to move all to configurations and localized, even the button labels.

src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java
mrsarm: Yes it is something discussed in the ticket, in the current master version if you regret to allow / deny the locations, you have the same problem, but it is true that it is more convenient just to remove the permission or the rejection from the app info in the Android settings than wipe all the data in order to clean this setting, but either case the user needs to go to the Android settings to do so.

> Can we instead check that somehow and only show the dialog if it's not already set?

It works like that, first it checks whether the user already allowed / denied the locations permissions, and only shows the dialog if the user never did. Users that already has the app installed and already allowed the app to access the location won't see the new dialog when upgrade to this version, and the app won't store any configuration.

Anyway, I'll check later if there is a way that in case the user rejects the permission request from our dialog, store the rejection in the same permissions settings that Android uses for that purpose, but because it is a rejection, not sure whether is possible.

build.gradle
mrsarm: OK, I'll leave it in a new line commented with a note that you have to enable it in order to test the app in a virtual device with Android 10 locally

n-orlowski:
@garethbowen I think should try to be as crystal clear as possible with copy and give users a heads up on the next screen (as the 2 screens back to back might be confusing):

Title: "Location Access"
Body: "{This app} collects location data when you submit a form to analyze and improve health outcomes in your area. Select **"Allow while using the app"** on the next screen to turn on your location access.
Ok button: "OK"
Deny button: "No thanks"

src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java
mrsarm: Back to this, I pushed a fix that was causing the form to be blocked when loc permissions were denied, and now testing again, turns out the user does not need to wipe the app data to rollback its decision to allow or deny the access to the location, it only needs to do the same that it does with the current master version: go to the app settings in the android settings and just add again or revoke the location permission, but no need to wipe the app data because the variable stored here is read after check the permissions against the Android API.

The only difference is that in case the user rejected the location permissions from the custom dialog instead of from the Android dialog, in the app settings you won't see the location permission accepted or rejected, the permission will be displayed like it was never accepted or rejected, and from there the user will be able to add the permission to access location in which case this variable stored will be ignored by the app.

Still will check later whether I can ask Android to store the negative of location access using the API like it was rejected from the native location dialog, but may be not possible.

build.gradle
mrsarm: Done.

mrsarm:
@garethbowen I have replaced the dialog by an activity / layout following the mockup provided in the issue (I updated the recording in the description). I couldn't find a way to avoid the rejection storage in the shared preferences, but as mentioned, the user can go to the app preferences in the Android settings and add the permission like it would been rejected from the Android dialog and the geolocation will work again.

@craig-landry don't worry about what I said in the meeting about calling methods of an activity from other activities, I was using the intents the wrong way, but I did a refactor and I think now is fine (without doing that).

src/main/res/layout/request_permission.xml
```diff
@@ -0,0 +1,77 @@
...
+    android:layout_height="wrap_content"
+    android:orientation="vertical"
+    android:paddingLeft="16dp"
+    android:paddingRight="16dp">
+
+    <ImageView
+        android:id="@+id/locIcon"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="128dp"
```
garethbowen, line 14: Have you tried this on a small screen?

src/main/res/values/strings.xml
```diff
@@ -45,4 +45,11 @@
 	<string name="promptChooseImage">Choose image</string>
 
 	<string name="spnCompressingImage">Compressing image…</string>
+
+	<string name="locRequest_title">Location Access</string>
+	<string name="locRequest_message">%s collects location data when you submit a form to analyze and improve health outcomes in your area. Select \"Allow while using the app\" on the next screen to turn on your location access.</string>
+	<string name="locRequest_okButton">Turn on</string>
+	<string name="locRequest_denyButton">No thanks</string>
+    <string name="locRequest_IconDescription">My location icon</string>
```
garethbowen, line 53: Nit: whitespace.

src/main/res/values/strings.xml
```diff
@@ -45,4 +45,11 @@
 	<string name="promptChooseImage">Choose image</string>
 
 	<string name="spnCompressingImage">Compressing image…</string>
+
+	<string name="locRequest_title">Location Access</string>
```
garethbowen, line 49: Use sentence case, not title case.

```suggestion
	<string name="locRequest_title">Location access</string>
```

src/main/res/values/strings.xml
```diff
@@ -45,4 +45,11 @@
...
+	<string name="locRequest_message">%s collects location data when you submit a form to analyze and improve health outcomes in your area. Select \"Allow while using the app\" on the next screen to turn on your location access.</string>
+	<string name="locRequest_okButton">Turn on</string>
+	<string name="locRequest_denyButton">No thanks</string>
+    <string name="locRequest_IconDescription">My location icon</string>
+	<string name="locRequest_mapIconDescription">Location in map icon</string>
```
garethbowen, line 54: I think the mix of snake and camel case for the names is weird. I know there's examples of this above, but let's start using camel case only.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -180,6 +183,25 @@ protected void onStart() {
...
+							this,
+							LOCATION_PERMISSIONS,
+							ACCESS_FINE_LOCATION_PERMISSION_REQUEST);
+				} else if (resultCode == RESULT_CANCELED) {		// user rejected
+					try {
+						this.locationRequestResolved();
+						settings.setUserDeniedGeolocation();
+					} catch (SettingsException e) {
+						error(e, "Error recording negative to access location");
+					} catch (Exception e) {
```
garethbowen, line 201: It's bad form to catch all Exceptions and I don't think it's required here. Can you remove this block?

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -298,18 +320,25 @@ private void setUpUiClient(WebView container) {
 				return true;
 			}
 			@Override public void onGeolocationPermissionsShowPrompt(final String origin, final GeolocationPermissions.Callback callback) {
+				trace(this, "onGeolocationPermissionsShowPrompt() :: origin=%s, callback=%s", origin, callback);
```
garethbowen, line 323: The callback doesn't have a nice toString so you end up just getting a memory reference which isn't very helpful. Just remove that from the trace to keep the logs more readable.

src/xwalk/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -51,6 +51,10 @@
 
 	private final static int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)Math.random();
 
+	private static final int DISCLOSURE_LOCATION_PERMISSION_REQUEST = 1122331;
```
garethbowen, line 54: I don't think this should be the same int as the one above. It shouldn't be a problem but if both classes exist then the wrong one may catch the response. I think it should be random, like the ACCESS_FINE_LOCATION_PERMISSION_REQUEST above.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -44,8 +44,11 @@
 	private static final int NON_SIMPRINTS_FLAGS = 0x7;
 	static final int GRAB_PHOTO = (0 << 3) | NON_SIMPRINTS_FLAGS;
 	static final int GRAB_MRDT_PHOTO = (1 << 3) | NON_SIMPRINTS_FLAGS;
+	private static final int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
 
-	private final static int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
+	private static final int DISCLOSURE_LOCATION_PERMISSION_REQUEST = 1122331;
```
garethbowen, line 49: I think this should be random as above.

garethbowen (changes requested):
Nice one! I've left a few minor comments inline...

src/main/res/values/strings.xml
mrsarm: Yes, I had the same filling but I was following the pattern. I'll change to camelCase only.

src/xwalk/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: I saw that the constant above that is assigned with a random number and looking in the internet why we would use a random number in a constant I couldn't find a good reason, but yes random numbers are used in `requestCode` variables when you need a new value each time, but this is not the case, the variable is only generated one time (`final`), so why we are doing that?

Constant are constants, and using a random value at startup time looks like a bad idea specially in Java, because you cannot use them in `switch` statements or annotations if the value set is assigned dynamically. Also there is the risk (unlikely) for collisions is two constants defined like that are caught in the `onActivityResult(..)` method that catch most events.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Yes same than above, I'll change it but I would like to understand what we do that.

src/main/res/layout/request_permission.xml
mrsarm: Yes, the recording in the description was made with a 5" display. Here is the same view in a 3.7" display:

Location Disclosure 3 4 inch

And here in a large one, 6.7":

Location Disclosure 6 7 inch

In all cases the components are displayed a bit different but looks acceptable I think. Is there a better way to define the margin othen than `dp` and other than define the same layout for each screen size? (unfortunately % is not supported in this layout class and other layout classes that support it require API 23 or above).

Taking into account that most of the deployments may use smaller sizes than 5" I can reduce the margin, maybe to 100dp?

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Yes I see the callback `toString` method is the `Object` implementation that only gets you the object id, and the origin string is always the URL set by the user so doesn't make sense.

src/xwalk/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Checking around the code again, I found that while in the Webview version we define the constant randomly like this:

```java
final static int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
```

In the XWalk version is:

```java
final static int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)Math.random();
```

`Math.random()` always return a `double` between 0 and 1 but never 1, so casting the result to `int` always get `0`. I guess this is a bug (but not resulting in errors for now) and not intended.

src/main/res/layout/request_permission.xml
mrsarm: Reduced to 100dp.

src/main/res/values/strings.xml
mrsarm: Changed all new string definitions to just camelCase.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Removed.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Removed.

src/main/res/layout/request_permission.xml
garethbowen: Awesome, that's perfect!

src/xwalk/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
garethbowen: Yeah definitely a bug - I should have fixed that when I fixed the Webview version, sorry!

I think the reason it's random is so it's different every time the app starts so you don't get a random response from a previous request. I'm happy to go with whatever the best practice is, if you can find one!

It seems like a strange API for someone used to callbacks...

garethbowen (approved):
Great work and a really impressive turn around time.

Feel free to make further tweaks as needed. I've pre-approved this so you're not blocked.

src/xwalk/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: :thinking: didn't know that the app can restart and still receive a response from a request from a previous process, in that case makes sense. Next event variable caught by the same method can be defined using different random ranges to avoid collisions, eg:

```
jshell> (int)(Math.random() * 1000)
$9 ==> 180

jshell> (int)(Math.random() * 1000)
$10 ==> 739

jshell> (int)(Math.random() * 1000) + 1000    // different range
$11 ==> 1819

jshell> (int)(Math.random() * 1000) + 1000
$12 ==> 1679

jshell>
```

So going to make it random and fix the one with the random that is always 0.

mrsarm: Fixed using random values,  and tested in XWalk and Webview

mrsarm:
Created tag: https://github.com/medic/medic-android/releases/tag/v0.7.0-alpha.1

src/main/res/values/strings.xml
```diff
@@ -45,4 +45,11 @@
 	<string name="promptChooseImage">Choose image</string>
 
 	<string name="spnCompressingImage">Compressing image…</string>
+
+	<string name="locRequestTitle">Location access</string>
+	<string name="locRequestMessage">%s collects location data when you submit a form to analyze and improve health outcomes in your area. Select \"Allow only while using the app\" on the next screen to turn on your location access.</string>
```
abbyad, line 50: The user is not necesarily aware of what consists and doesn't consist of a form, so it would be clearer to avoid mentioning forms here.

```suggestion
	<string name="locRequestMessage">%s collects location data to analyze and improve health outcomes in your area. Select \"Allow only while using the app\" on the next screen to turn on your location access.</string>
```

README.md
```diff
@@ -5,6 +5,17 @@ The medic-android application is a thin wrapper to load the [CHT Core Framework]
...
 
+## 0.7.0
+
+### Changes
+
+- [feature] [medic-android#136](https://github.com/medic/medic-android/issues/136): Add UI for prominent disclosure when requesting for permissions.
+
+### Notes
+
+The new view that requests location permissions has in the source code text translated to the main languages we support. These translation don't follow the rules from the [How To Manage Translations](https://docs.communityhealthtoolkit.org/core/overview/translations/) because Android has its own [localization framework](https://developer.android.com/guide/topics/resources/localization) and there are just a handful of phrases translated.
```
abbyad, line 16: Good to have this! Suggested changes for clarity:

```suggestion
The text used in the new location permission request is in the Android wrapper app itself (`medic-android`), and translated differently than [CHT Core labels](https://docs.communityhealthtoolkit.org/core/overview/translations/). Any additions or modifications to translations in `medic-android` are done in the `strings.xml` files according to the [Android localization framework](https://developer.android.com/guide/topics/resources/localization).
```

abbyad (changes requested):
Looks good! Minor comments and suggestions for the release notes, but should not be blocking as the PR is pre-approved.

abbyad (approved):
I take it back, "request changes" did in fact block despite the other approval.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -44,8 +44,13 @@
...
 	static final int GRAB_PHOTO = (0 << 3) | NON_SIMPRINTS_FLAGS;
 	static final int GRAB_MRDT_PHOTO = (1 << 3) | NON_SIMPRINTS_FLAGS;
+	private static final int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
 
-	private final static int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
+	// Use different ranges of random values for other request codes caught by the same method,
+	// eg. (int)(Math.random() * 1000) + 1000
+	private static final int DISCLOSURE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
+
+	private static final String[] LOCATION_PERMISSIONS = { Manifest.permission.ACCESS_FINE_LOCATION };
```
craig-landry, line 53: The `Math.random()` here worries me. These codes should be fixed. As it is now, this number will be random (at class init time) between 0 and 999 (inclusive). There is a chance that this will match down on line 176 and go into the photo branch and not end up in the new permissions branch (line 190) as desired.

From our testing I think we've seen some very rare cases where the secondary permissions modal did not show, which would make sense since that is triggered in that new conditional branch, which wouldn't get hit if we got unlucky with the randomization. Also I know it's been reported that when the app is restarted (versus staying running when switching "windows" in Android) that has led to differing behavior, which would also line up with this randomization changing at class loading time.

This new constant needs to be a fixed integer, but we also need to do the same for the other random constant here to avoid that overlapping with our newly selected constant.

craig-landry (changes requested):

eaaliprantis:
Hi.

I have a suggestion to make.

For dealing with permissions, the first thing that when the app opens is to request all of the permissions right away. I’ve published apps before to do that. And if you do not have that done right away, you don’t get approved in the App Store.

Just food for thought

craig-landry:
> Hi. I have a suggestion to make. For dealing with permissions, the first thing that when the app opens is to request all of the permissions right away. I’ve published apps before to do that. And if you do not have that done right away, you don’t get approved in the App Store. Just food for thought
> […](#)
> Sent from my iPhone
> On Jan 8, 2021, at 10:56 AM, Craig Landry ***@***.***> wrote: ﻿ @craig-landry requested changes on this pull request. In src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java: > + private static final int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000); - private final static int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000); + // Use different ranges of random values for other request codes caught by the same method, + // eg. (int)(Math.random() * 1000) + 1000 + private static final int DISCLOSURE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000); + + private static final String[] LOCATION_PERMISSIONS = { Manifest.permission.ACCESS_FINE_LOCATION }; The Math.random() here worries me. These codes should be fixed. As it is now, this number will be random (at class init time) between 0 and 999 (inclusive). There is a chance that this will match down on line 176 and go into the photo branch and not end up in the new permissions branch (line 190) as desired. From our testing I think we've seen some very rare cases where the secondary permissions modal did not show, which would make sense since that is triggered in that new conditional branch, which wouldn't get hit if we got unlucky with the randomization. Also I know it's been reported that when the app is restarted (versus staying running when switching "windows" in Android) that has led to differing behavior, which would also line up with this randomization changing at class loading time. This new constant needs to be a fixed integer, but we also need to do the same for the other random constant here to avoid that overlapping with our newly selected constant. — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Thanks for your thoughts on this @eaaliprantis but that won't work for us here. We are purposefully taking a different approach here as Google requires the prompt to happen in the context of the GPS action taking place. It would definitely be easier to ask for everything up front, but we are not permitted to handle it that way for this app.

eaaliprantis:
Interesting that Google changes their permission based on GPS actions instead of being done initially at the beginning

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
craig-landry: I've put together https://github.com/medic/medic-android/pull/146 to address the result code collisions.

abbyad:
> Interesting that Google changes their permission based on GPS actions instead of being done initially at the beginning

Ya, it seems to be part of an effort to avoid abuses of privacy, and for users to be aware when and why permissions are needed. The [current recommendation](https://developer.android.com/training/permissions/requesting#allow-system-manage-request-code) is to _"ask for permissions in context, when the user starts to interact with the feature that requires it."_

craig-landry (approved):

PR #146, opened 2021-01-08
Avoid result code collisions

craig-landry:
The new code to handle the location permissions risked colliding with the scheme used for handling simprints result codes. This change sets the new result code to a constant in the reserved space.

The prior version also used random numbers, which is removed as a side effect of this specific code selection.

This should address the intermittent issues we have seen during testing, which I believe were a result of the new intent returning and having the undesired effect of being handled by the photo-grabbing handler code.

### DIFF
--- a/src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
+++ b/src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
@@ -49,12 +49,12 @@ public class EmbeddedBrowserActivity extends LockableActivity {
 	 * bitmask is used to respect the scheme on that side of things.
 	 * */
 	private static final int NON_SIMPRINTS_FLAGS = 0x7;
-	static final int GRAB_PHOTO = (0 << 3) | NON_SIMPRINTS_FLAGS;
-	static final int GRAB_MRDT_PHOTO = (1 << 3) | NON_SIMPRINTS_FLAGS;
-	static final int DISCLOSURE_LOCATION_PERMISSION_REQUEST = (2 << 3) | NON_SIMPRINTS_FLAGS;
+	static final int GRAB_PHOTO_ACTIVITY_REQUEST_CODE = (0 << 3) | NON_SIMPRINTS_FLAGS;
+	static final int GRAB_MRDT_PHOTO_ACTIVITY_REQUEST_CODE = (1 << 3) | NON_SIMPRINTS_FLAGS;
+	static final int DISCLOSURE_LOCATION_ACTIVITY_REQUEST_CODE = (2 << 3) | NON_SIMPRINTS_FLAGS;
 
 	// Arbitrarily selected value
-	private static final int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = 7038678;
+	private static final int ACCESS_FINE_LOCATION_PERMISSION_REQUEST_CODE = 7038678;
 
 	private static final String[] LOCATION_PERMISSIONS = { Manifest.permission.ACCESS_FINE_LOCATION };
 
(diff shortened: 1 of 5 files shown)


craig-landry:
@garethbowen  I'd appreciate a detailed review of this code. I left most of the logic alone and the only real change here is to get this new logic integrated into that photo/simprints handler router (aka case statement). Prior to digging in here I was not familiar with the request code scheme seen here, which is where I am eager to have others verify my assumptions. As I can tell, that matches up with the CHT webapp logic found in webapp/src/js/services/simprints.js:22

cc @abbyad

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -40,15 +40,21 @@
...
 	static final int GRAB_PHOTO = (0 << 3) | NON_SIMPRINTS_FLAGS;
 	static final int GRAB_MRDT_PHOTO = (1 << 3) | NON_SIMPRINTS_FLAGS;
-	private static final int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
+	static final int DISCLOSURE_LOCATION_PERMISSION_REQUEST = (2 << 3) | NON_SIMPRINTS_FLAGS;
 
-	// Use different ranges of random values for other request codes caught by the same method,
-	// eg. (int)(Math.random() * 1000) + 1000
-	private static final int DISCLOSURE_LOCATION_PERMISSION_REQUEST = (int)(Math.random() * 1000);
+	// Arbitrarily selected value
+	private static final int ACCESS_FINE_LOCATION_PERMISSION_REQUEST = 7038678;
```
garethbowen, line 57: I think some of the confusion here is that these codes look kind of similar but they are used in different ways. The first three are used for activities and the last one is used for a permission request. In particular `DISCLOSURE_LOCATION_PERMISSION_REQUEST` should be renamed because it's _not_ a permission request, it's launching an activity. This will help give a clue as to why the first three need to be masked to not conflict with simprints but the last one doesn't, and hopefully help this not happen again!

I think they should be renamed as follows:

`GRAB_PHOTO` -> `GRAB_PHOTO_ACTIVITY_REQUEST_CODE`
`GRAB_MRDT_PHOTO` -> `GRAB_MRDT_PHOTO_ACTIVITY_REQUEST_CODE`
`DISCLOSURE_LOCATION_PERMISSION_REQUEST` -> `DISCLOSURE_LOCATION_ACTIVITY_REQUEST_CODE`
`ACCESS_FINE_LOCATION_PERMISSION_REQUEST` -> `ACCESS_FINE_LOCATION_PERMISSION_REQUEST_CODE`

garethbowen (changes requested):
Great work!

I've left a comment that I think would make this more readable, which of course applies to both versions of the class.

I never managed to find any documentation about what the request code should be so your arbitrarily picked value works for me!

craig-landry:
Thanks for looking @garethbowen. I agree with your take on the codes and have updated the names to match your advice.

garethbowen (approved):
Nice one! Thanks.

garethbowen:
@craig-landry I'll leave it up to you to decide if this should be ATed before or after merging into `136-request-location-disclosure`... either is fine for me.

craig-landry:
> @craig-landry I'll leave it up to you to decide if this should be ATed before or after merging into `136-request-location-disclosure`... either is fine for me.



> @craig-landry I'll leave it up to you to decide if this should be ATed before or after merging into `136-request-location-disclosure`... either is fine for me.

Thanks @garethbowen. We should test this, but since we'll need to test after merging anyway, I'm going to merge this into @mrsarm's branch now and we can test the final product that way.

cc @ngaruko

PR #149, opened 2021-01-18
Add safari doctors flavour

derickl:


mrsarm (changes requested):
Nice! I have one request and another that not sure if needed:

- Please remove the `src/safaridoctors_kenya/res/.DS_Store` file and rebase to not keep it in the git log
- All flavors are configured in the `productFlavors` section in Gradle [here](https://github.com/medic/medic-android/blob/master/build.gradle#L160), I think you should add it but not sure the impact of doing so.

mrsarm:
> * All flavors are configured in the `productFlavors` section in Gradle [here](https://github.com/medic/medic-android/blob/master/build.gradle#L160), I think you should add it but not sure the impact of doing so.

Forget about this comment, not sure why I didn't see it when reviewing :facepalm:

derickl:
Closing this since I messed up the branch history

PR #152, opened 2021-01-21
Add safaridoctors flavour

derickl:
Supersedes #149 

All feedback has been incorporated


mrsarm (approved):
LGTM.

Please after merge into master instead of release an alpha with these changes let me rebase this [PR](https://github.com/medic/medic-android/pull/150) to add `safaridoctors_kenya` to the list of flavors that don't use the storage perm, and then we can release the 0.7.2-alpha.XX with this new flavor (this flavor won't use storage perm, right?) because actually the access to the storage is not working in the app right now.

derickl:
It doesn't need that permission

derickl:
Back to you @mrsarm

Issue #148, opened 2021-01-15
Prominent disclosure when "handling users' Files"


kennsippell:
Follow-up after https://github.com/medic/medic-android/issues/136

Apps are getting denied from the app store because we are requesting `READ_EXTERNAL_STORAGE` without a prominent alert asking for access.  See screenshot for details form google. 

![Screenshot from 2021-01-15 15-21-20](https://user-images.githubusercontent.com/9014751/104787789-5b3cc000-5745-11eb-9c1e-6f3af65db346.png)

mrsarm:
Created the [PR#150](https://github.com/medic/medic-android/pull/150) where the `READ_EXTERNAL_STORAGE` permission is removed from selected flavors.

Tomorrow I'll check the necessary changes to implement at runtime the prominent disclosure requested by Google, but it's possible the move may require changes in the way the CHT requires access to the file storage (though thinking it well, it shouldn't, because we are not accessing any JS API to request storage access, it is required when the user clicks in an upload field :thinking: )

So, we have 2 alternatives:

- **The changes in the Android App requires changes in the CHT**: in this case, we need to release a new version of the CHT, and upgrade all the deployments currently have the app published in the Play Store. In this case, the PR mentioned above allows us to skip the CHT upgrade for those deployments that really don't use upload fields, therefore don't need the storage location in the app.
- **The changes in the Android App DON'T require changes in the CHT**: we can work in a new app release without the need to change anything in the CHT, but still we will need to publish a new version in the Play Store with the changes, and they aren't backward compatible with older versions of the App, specifically v0.5.0 where we change the way we request location access. 

Regardless of the choice, we can still keep the changes from  [PR#150](https://github.com/medic/medic-android/pull/150) and remove the storage permission at all for those flavors, making the Play Store approval process simpler, but we will need to keep in sync what flavors do access to the local storage to avoid the app to crash. In this case we can also add a change using the Android API to check whether the app has the permission in the manifest and avoid the app crashing and letting the user know about the error, despite it shouldn't happen.

mrjones-plip:
Checking in with existing deployments, I asked @derickl:

> are any  deployments using file uploads?

To which he replied:

> None that I know of

@derickl  - I'm going to assume it's OK if we remove the feature as needed for this  ticket. We can re-introduce the feature as needed at a later date which can then be play store compliant.

CC @abbyad @michaelkohn @craig-landry

mrjones-plip:
[...]

PR #156, opened 2021-01-28
Add release notes for v0.7.3

mrsarm:
Release notes.

Issue: #148

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,18 @@ The medic-android application is a thin wrapper to load the [CHT Core Framework]
 
 # Release notes
 
+### 0.7.3
+
+### Changes
+
+- [improvement] [medic-android#148](https://github.com/medic/medic-android/issues/148): Remove storage access for most flavors that don't use the permission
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,18 @@ The medic-android application is a thin wrapper to load the [CHT Core Framework]
 
 # Release notes
 
+### 0.7.3
+
```
garethbowen, line 9: Are there any breaking changes or upgrade steps people should know about? For example, what if they were using the permission before?

Should it be version 0.8.0 to make that more clear?

mrsarm: We only have removed the permission from the flavors we confirmed are not using the permission, while the "unbranded" flavor used for general testing still have the perm, despite it doesn't look to work as #127 points out. So no, it won't break any flavor.

garethbowen: Awesome. Thanks for clarifying!

garethbowen (approved):
LGTM!

PR #160, opened 2021-02-12
Add Gem TripleEighty Apk Flavour

billwambua:
REF https://github.com/medic/medic-projects/issues/7807


billwambua:
We don't need the APK since we will not be using the GemTripleEighty instance.

PR #162, opened 2021-03-11
134 - Improve UX of crosswalk to webview migration

mrsarm:
Improve the UX after the migration restarting the app.

Issue: https://github.com/medic/medic-android/issues/134

### TO DO:

- [x] Restart the app after the migration to avoid the login page
- [x] Add a nice activity / layout following the mockups while the user wait for the migration
- [x] ~Add also a better UI with activities / layout when the connection is lost~ → Moved to https://github.com/medic/medic-android/issues/163

### Other improvements
- Improved logging in the app (sorry if it adds noise, I started just with a few corrections and then I couldn't stop fixing it).
- Upgrade build library com.android.tools.build:gradle ~4.1.1~ → 4.1.2, Android Studio marks the older version as insecure.

Medic Android Migration Fixed 2021-03-12 16-01

### DIFF
--- a/src/main/res/values-fr/strings.xml
+++ b/src/main/res/values-fr/strings.xml
@@ -6,7 +6,7 @@
 	<string name="locRequestDenyButton">Non merci</string>
 	<string name="locRequestIconDescription">Icône de mon emplacement</string>
 
-	<string name="loadingIconDescription">Chargement…</string>
-	<string name="upgradingTitle">Mettre à jour votre application</string>
-	<string name="upgradingMessage">Vous passez à la dernière version. Veuillez patienter pendant la migration de vos données.</string>
+	<string name="loadingIconDescription">Chargement en cours…</string>
+	<string name="upgradingTitle">Mise à jour de votre application</string>
+	<string name="upgradingMessage">Vous passez à la nouvelle version. Veuillez patienter pendant la preparation de vos données.</string>
 </resources>
--- a/src/main/res/values-fr/strings.xml
+++ b/src/main/res/values-fr/strings.xml
@@ -5,4 +5,8 @@
 	<string name="locRequestOkButton">Activer</string>
 	<string name="locRequestDenyButton">Non merci</string>
 	<string name="locRequestIconDescription">Icône de mon emplacement</string>
+
+	<string name="loadingIconDescription">Chargement…</string>
+	<string name="upgradingTitle">Mettre à jour votre application</string>
+	<string name="upgradingMessage">Vous passez à la dernière version. Veuillez patienter pendant la migration de vos données.</string>
 </resources>
--- a/src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
+++ b/src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
@@ -301,7 +301,7 @@ public class EmbeddedBrowserActivity extends LockableActivity {
 
 	private void browseTo(Uri url) {
 		String urlToLoad = getUrlToLoad(url);
-		if(DEBUG) trace(this, "Pointing browser to %s", redactUrl(urlToLoad));
+		trace(this, "Pointing browser to: %s", redactUrl(urlToLoad));
 		container.loadUrl(urlToLoad, null);
 	}
 
@@ -323,7 +323,7 @@ public class EmbeddedBrowserActivity extends LockableActivity {
 				return true;
 			}
 			@Override public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, WebChromeClient.FileChooserParams fileChooserParams){
-				if(DEBUG) trace(this, "onShowFileChooser() :: %s,%s,%s", webView, filePathCallback, fileChooserParams);
+				trace(this, "onShowFileChooser() :: webView: %s,filePathCallback: %s,fileChooserParams: %s", webView, filePathCallback, fileChooserParams);
 
 				boolean capture = fileChooserParams.isCaptureEnabled();
 				trace(this, "onShowFileChooser() capture :: %s", capture);
@@ -417,8 +417,9 @@ public class EmbeddedBrowserActivity extends LockableActivity {
 			@Override
 			public void onReceivedError(WebView view, WebResourceRequest request, WebResourceError error) {
 				String failingUrl = request.getUrl().toString();
-				if(DEBUG) trace(this, "onReceivedLoadError() :: %s,%s,%s", failingUrl, error.getErrorCode(), error.getDescription());
-				if(!getRootUrl().equals(failingUrl)) {
+				trace(this, "onReceivedLoadError() :: url: %s, error code: %s, description: %s",
+						failingUrl, error.getErrorCode(), error.getDescription());
+				if (!getRootUrl().equals(failingUrl)) {
 					super.onReceivedError(view, request, error);
 				} else {
 					evaluateJavascript(String.format(
--- a/src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
+++ b/src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
@@ -40,6 +40,7 @@ import static org.medicmobile.webapp.mobile.MedicLog.warn;
 import static org.medicmobile.webapp.mobile.SimpleJsonClient2.redactUrl;
 import static org.medicmobile.webapp.mobile.Utils.createUseragentFrom;
 import static org.medicmobile.webapp.mobile.Utils.isUrlRelated;
+import static org.medicmobile.webapp.mobile.Utils.restartApp;
 
 @SuppressWarnings({ "PMD.GodClass", "PMD.TooManyMethods" })
 public class EmbeddedBrowserActivity extends LockableActivity {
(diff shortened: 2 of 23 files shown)


src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -435,6 +450,26 @@ public void onReceivedError(WebView view, WebResourceRequest request, WebResourc
 							"}", error.getErrorCode(), error.getDescription()), false);
 				}
 			}
+
+			@Override
+			public void onPageFinished(WebView view, String url) {
+				trace(this, "onPageFinished() :: url: %s, isMigrationRunning: %s", url, isMigrationRunning);
+				if (isMigrationRunning && url.contains("/login")) {
+					isMigrationRunning = false;
```
garethbowen, line 458: If the url does not contain `/login` that means the cookie has been found, right? If so, I think it makes sense to set `isMigrationRunning = false` so we never have to check the URL again, eg:

```
				if (isMigrationRunning) {
					isMigrationRunning = false;
					 if (url.contains("/login")) {
```

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -78,6 +80,8 @@ public void onReceiveValue(String result) {
 	private PhotoGrabber photoGrabber;
 	private SmsSender smsSender;
 
+	private boolean isMigrationRunning = false;
```
garethbowen, line 83: With this as a boolean it'll reset to false on restart. This may be an issue if it crashes or the user closes it. Can we store it somewhere persistent instead?

garethbowen (commented):
I know you hadn't marked this for review, but thought it might help to add my two cents earlier rather than later!

I've added a couple of suggestions inline.

Additionally, the `if (DEBUG)` statements are there as a slight performance boost so in production the app doesn't have to initialise and format the parameters only for them to be GCed again almost immediately. This is best practice as far as I'm aware.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Yes, but solve that means make the migration process more resilient, with the ability to resume in the cases you mention, but it is out of scope I think for this ticket that only intent to improve the UI so users are not presented to the login page which cause many problems described in the ticket.

Currently, If the app crashes but the migration was finished, there won't be a problem, because the app will be restarted and with our without this patch the UI will load the CHT dashboard because a that point the cookies will be available again. If the crash happens during the migration, is likely the data will be corrupted and the user will need to reinstall the app losing the data, but again this PR cannot address that now.

mrsarm:
> Additionally, the `if (DEBUG)` statements are there as a slight performance boost so in production the app doesn't have to initialise and format the parameters only for them to be GCed again almost immediately. This is best practice as far as I'm aware.

@garethbowen yes actually I removed it not just to make the code more readable without so many `if (DEBUG)` statements but also because performance, due that the `trace` method already do that check, and it does not perform the string formatting if `DEBUG` is `true`. I only kept the `if (DEBUG)` in the cases where the formatting is done outside the `trace` method because it requires more complex concatenations.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: This method is called with the flag `isMigrationRunning = true` with other URLs than `/login` before, so changing that would cause to set the variable `false` before reach the point desired.

mrsarm:
Ready for review again, comments about the changes here: https://github.com/medic/medic-android/issues/134#issuecomment-797729960

mrsarm:
### Quick test instructions

Not yet in AT but here are some observations of how to test the PR.

Because the XWalk version does not work in Android >= 10, and the Webview version is compiled to work only in Android >= 10 for now in the master branch and this branch, the way to test the PR would be install the XWalk version in an Android <10 phone, then upgrade the phone  to Android >=10 and finally install the version on this PR, which is is too much for testing, and may not be an option because the device used for testing does not support Android upgrades to the version 10.

What I do for test is:

1. Get a phone with Android <10 and >=5 (Webview doesn't work with Android 4.4).
1. Install in the phone the unbranded XWalk version of the app using the latest available from our releases: [armeabi-v7a](https://github.com/medic/medic-android/releases/download/v0.7.3/medic-android-v0.7.3-unbranded-xwalk-armeabi-v7a-release.apk) or [arm64-v8a](https://github.com/medic/medic-android/releases/download/v0.7.3/medic-android-v0.7.3-unbranded-xwalk-arm64-v8a-release.apk) APK.
2. Setup the app with one of the CHT environments suggested by the app, I could reproduce the error easily with the Gamma dev environment.
3. Log in with an offline user to sync the app and some user data.
4. Then to be able to install the Webview version of this PR over the version installed in your phone, you need to downgrade the minimal SDK supported in your local development installation of this branch, so go to [build.gradle#L164](https://github.com/medic/medic-android/blob/5fec28df37865b75c636e6585fe5c5d0de361884/build.gradle#L164) and set a API version supported by your phone or below, eg.:

   ```groovy
   minSdkVersion 23     // Android 6
   ```

5. Run the `unbrandedWebviewDebug` flavor of the app of this branch with `adb`, `gradle` or Android Studio over the same phone where the Xwalk version was installed.
6. You will see the app starts, and then after a few seconds the splash screen with the upgrade notification while the migration process happens, as described [here](https://github.com/medic/medic-android/issues/134#issuecomment-797729960) in the ticket. Then instead of being presented with the login page of the CHT you should see the normal dashboard of the app, with the data already synchronized there.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
garethbowen: I'm worried about edge cases like... What if the user spawns a [second instance](https://github.com/medic/medic-android/issues/122) of this Activity? What happens if two migrations run in parallel, or one after the other?

However adding some sort of persistent flag is also risky... what if it gets set to `migrating = true` and then the migration fails? It may never recover.

This is going to have to be tested really thoroughly to avoid bricking or corrupting phones. Are you confident this is the safest approach?

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -435,6 +451,26 @@ public void onReceivedError(WebView view, WebResourceRequest request, WebResourc
...
+					CookieManager cookieManager = CookieManager.getInstance();
+					String cookie = cookieManager.getCookie(appUrl);
+					if (cookie == null) {
+						log(this, "onPageStarted() :: Migration process in progress, and " +
+								"cookies were not loaded, restarting ...");
+						Context context = view.getContext();
+						Intent intent = new Intent(context, StartupActivity.class);
+						intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+						context.startActivity(intent);
+						Runtime.getRuntime().exit(0);
```
garethbowen, line 469: This section is really important, and quite difficult to understand. Consider pulling out a private function that can be named clearly. In addition it probably needs some commenting about what this is doing and why it works.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -435,6 +451,26 @@ public void onReceivedError(WebView view, WebResourceRequest request, WebResourc
...
 			}
+
+			@Override
+			public void onPageStarted(WebView view, String url, Bitmap favicon) {
+				trace(this, "onPageStarted() :: url: %s, isMigrationRunning: %s", url, isMigrationRunning);
+				if (isMigrationRunning && url.contains("/login")) {
+					isMigrationRunning = false;
+					CookieManager cookieManager = CookieManager.getInstance();
+					String cookie = cookieManager.getCookie(appUrl);
+					if (cookie == null) {
```
garethbowen, line 462: What happens if a user's cookies have expired (or be wiped manually)? I think it'll work... migration will pick up the IndexedDB, then on restart `isMigrationRunning` will be false, and the user will go to the login page (as expected). Is that right? Have you tested this case?

garethbowen (commented):
The video walkthrough looks awesome. Great work!

I've left some comments inline. Because of the risk of data loss or bricking phones I'm being even more pedantic than usual...

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Yes, I tried to install this version over a XWalk installation that has loaded the webapp but logged out, and the migration runs and then the app display the login page as expected.

Also installed the app without previous installations of Xwalk to ensure the migration process is not started. Same if it was a XWalk installation but with no data previously loaded.

src/main/res/values-fr/strings.xml
```diff
@@ -5,4 +5,8 @@
 	<string name="locRequestOkButton">Activer</string>
 	<string name="locRequestDenyButton">Non merci</string>
 	<string name="locRequestIconDescription">Icône de mon emplacement</string>
+
+	<string name="loadingIconDescription">Chargement…</string>
+	<string name="upgradingTitle">Mettre à jour votre application</string>
+	<string name="upgradingMessage">Vous passez à la dernière version. Veuillez patienter pendant la migration de vos données.</string>
```
mrsarm, line 11: @abbyad could you please check whether these french translations are OK?

garethbowen (approved):
Awesome!

Approved - the build is failing but I'll let you dig in to that.

Feel free to create a separate issue/PR for the service worker cache so we can start the AT process on this one.

mrsarm:
> Awesome!
> 
> Approved - the build is failing but I'll let you dig in to that.

Thanks @garethbowen ! Yes the build error is due an issue in the jCenter servers that are responding with a HTTP 403 error when CI tries to pull one of the dependencies. I'll try to rebuild in a few minutes.

> Feel free to create a separate issue/PR for the service worker cache so we can start the AT process on this one.

Yes, going to do that to move forward with testing.

src/main/res/values-fr/strings.xml
abbyad: Pretty close! Minor changes suggested here:

```suggestion
	<string name="loadingIconDescription">Chargement en cours…</string>
	<string name="upgradingTitle">Mise à jour de votre application</string>
	<string name="upgradingMessage">Vous passez à la nouvelle version. Veuillez patienter pendant la preparation de vos données.</string>
```

PR #165, opened 2021-03-19
163 - Improve connection errors UX (Webview)

mrsarm:
Activity + Layout to catch connection errors and give the user the ability to retry after fix the connection, and the ability to leave the application only if it doesn't cause the user to lost data (when user is migrating the app).

Issue: https://github.com/medic/medic-android/issues/163

### Implementation

A new activity and layout were implemented to catch connection errors and avoid the not so friendly Chrome view error screen. The implementation only covers Webview (and therefore the migration from XWalk to Webview). I tried to use the same implementation in XWalk (though I needed to make some adaptations of how XWalk catch errors) and it worked, but for some weird error it cause the app to fail after restart if the user leaves the app with the error activity, so taking into account that is not a priority and we are going to drop XWalk soon, I left Xwalk error management as it is now.

### Improvements

- Spinner while the app retries the connection.
- Always show a button "Retry" so the user can retry after recover the connection. Previously the button was not displayed some times as show in the screenshot from the ticket.
- User can retry many times without problems. Previously retrying with the html button "Retry" many times caused the button to disappear at some point, some times even after the first try.
- Allow to close the app pressing back in the login workflow. In the previous implementation leaving the app and entering again caused the user to see again the connection error until force the full close of the app from the app manager of Android. Now because pressing back really close the app, the second time is opened the loading starts again.
- In the migration process the error activity does not allow the user to leave the app because testing that scenario in my device caused some times to lose the data, so instead the user sees a "toast" message saying that it needs to wait until the migration process finish.
- A button "More info" is provided to see the error code that caused the connection error.

#### Login page

Here is a recording with the user trying to load the app and sign-in for the first time. It shows different scenarios: user entering without connection, then retry again without connection. User press back living the app. Then enter again into the app, the app allows to sign-in without force closing the app when the connection is recovered.

Medic Android - No Internet - Login 2

#### Migration process

In the recording below the app was installed on top of the Xwalk version. It shows how the migration popup is shown and immediately the lack of connection is displayed by the connection error activity. The user then reconnects the WiFi and the process continue without problems.

Medic Android - No Internet - Migration running 2

Below the user tries to leave the app pressing back but the exit is prevented and a "toast" message is displayed asking the user to wait the migration process to finish:

Medic Android - No Internet - Migration running - Wait

Finally, here is the button "More info" showing the error code:

Medic Android - No Internet - More Info

### Notes

- The activity only intercepts connection errors when the app fails to load an URL (login) or reload the current page (migration process, or when the user press "Retry"). Ajax calls from the webapp are not intercepted, and are handled by the CHT webapp itself, nothing change there.
- When a page fails to load like the login page, the old Chrome error page is visible for a fraction of seconds until is hidden by the connection error activity on top, we cannot avoid that :disappointed: . 
- A phone with black theme was used for the recording (Android 6). I also tested the features with an Android 11 phone with normal theme and the new activity layout has white background with black text as the mockup from the ticket.
- To test the migration workflow, check the steps here: https://github.com/medic/medic-android/pull/162#issuecomment-798546578.

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/ClosableAppActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/ClosableAppActivity.java
@@ -3,75 +3,46 @@ package org.medicmobile.webapp.mobile;
 import android.app.Activity;
 import android.content.Intent;
 import android.os.Bundle;
-import android.os.Handler;
 import android.widget.Toast;
 
 /**
  * Base class for activities that when back is pressed
- * two times the app closes.
+ * it closes the app instead of move backward through
+ * the history of screens, or prevent the user to either
+ * close or go back in the activities stack.
  *
- * When the user press back for the first time, a message
- * defined in the `R.string.backToExit` string is shown
- * explaining that it has to press back one more time. The
- * message can be customized in the intent with the extended
- * data "backPressedMessage"
- * (calling `intent.putExtra("backPressedMessage", "...")`).
+ * By default when the user press back the app is exited.
  *
- * Also the opposite action can be configured: the app
+ * The opposite action can be configured: the app
  * not closing nor the activity when the user press
  * back one or more times: to do so the `Intent` used to launch
  * the activity needs to be set with the extended data "isClosable"
  * to `false` (calling `intent.putExtra("isClosable", false)`).
  * No message is shown to the user when back is pressed unless
  * the extended data "backPressedMessage" is also set with the
- * desired message.
+ * desired message string.
  */
 public abstract class ClosableAppActivity extends Activity {
 
-	private static final int WAIT_MILLIS = 2 * 1000;    // Back need to be pressed 2 times before this time
-
-	private boolean closeableApp = true;                // if true, back pressing 2 times close the app
-	private boolean backToExitPressedOnce = false;      // whether back was pressed the last WAIT_MILLIS
-	private String backPressedMessage = null;			// Customized message to show when back is pressed
+	private boolean isClosable = true;			// if true, back pressing close the app
+	private String backPressedMessage = null;	// Customized message to show when back is pressed
+												// and the app activity is not closable
 
 	@Override public void onBackPressed() {
-		if (closeableApp) {
-			if (backToExitPressedOnce) {
-				finishAffinity();
-				return;
-			}
-			backToExitPressedOnce = true;
-
-			toastMessage();
-
-			new Handler().postDelayed(new Runnable() {
-				@Override public void run() {
-					// After WAIT_MILLIS clean the back pressed history
-					backToExitPressedOnce = false;
-				}
-			}, WAIT_MILLIS);
+		if (isClosable) {
+			finishAffinity();
 		} else {
 			// Back pressed events won't close the app nor close the activity
-			toastMessage();
+			if (backPressedMessage != null) {
+				Toast.makeText(this, backPressedMessage, Toast.LENGTH_SHORT).show();
+			}
 		}
 	}
 
 	@Override public void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		Intent i = getIntent();
-		closeableApp = i.getBooleanExtra("isClosable", true);	// not closable (the default value is true)
-		backPressedMessage = i.getStringExtra("backPressedMessage");	// default null
-	}
-
-	private void toastMessage() {
-		if (closeableApp) {
-			if (backPressedMessage != null) {
-				Toast.makeText(this, backPressedMessage, Toast.LENGTH_SHORT).show();
-			} else {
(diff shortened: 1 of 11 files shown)


src/main/res/values-fr/strings.xml
```diff
@@ -9,4 +9,11 @@
 	<string name="loadingIconDescription">Chargement en cours…</string>
 	<string name="upgradingTitle">Mise à jour de votre application</string>
 	<string name="upgradingMessage">Vous passez à la nouvelle version. Veuillez patienter pendant la preparation de vos données.</string>
+
+	<string name="connErrorTitle">Pas de connexion Internet</string>
+	<string name="connErrorMessage">Vous ne semblez pas connecté à Internet. Veuillez vérifier votre connexion et réessayer.</string>
+	<string name="connErrorRetry">RECOMMENCEZ</string>
+
+	<string name="backToExit">Veuillez appuyer à nouveau pour quitter</string>
+	<string name="waitMigration">Veuillez attendre la fin du processus de migration</string>
```
mrsarm, line 18: @abbyad , I need your French skills again here. Also if you want to check the English strings would be OK, the first 3 were proposed by Nicole but the last 2 were added by me.

abbyad: It's not entirely clear what ["back" button](https://github.com/medic/medic-android/blob/aa7091dc25c7ab2064316158da9255087909508c/src/main/res/values/strings.xml#L63) is here, but some suggestions:

```suggestion
	<string name="connErrorTitle">Aucune connexion Internet</string>
	<string name="connErrorMessage">Vous n'êtes pas bien connecté à internet. Veuillez vérifier votre connexion et réessayer.</string>
	<string name="connErrorRetry">RÉESSAYER</string>

	<string name="backToExit">Veuillez appuyer à nouveau pour quitter</string>
	<string name="waitMigration">Veuillez attendre que la migration de données termine</string>
```

src/main/java/org/medicmobile/webapp/mobile/Utils.java
```diff
@@ -77,4 +79,15 @@ static void restartApp(Context context) {
...
 	}
+
+	static boolean isConnectionError(WebResourceError error) {
+		switch (error.getErrorCode()) {
+			case ERROR_HOST_LOOKUP:
+			case ERROR_PROXY_AUTHENTICATION:
+			case ERROR_CONNECT:
+			case ERROR_TIMEOUT:
+				return true;
+		}
```
garethbowen, line 90: I think this is the right collection of errors to catch. There are some others that are maybes for me, but this is fine for now. One concern I have is it'll be hard to check which of these errors a user is getting without checking the log. How do you feel about adding a "more information" button or similar to reveal which error code you got?

garethbowen (changes requested):
Very nice! A couple of comments inline, but this is an awesome improvement.

A couple of additional questions:

1. Why add the double back to exit feature? Is it just to block the default behaviour of the back button? I think it'd be cleaner to let a single back press behave normally unless migrating.
2. I'm concerned that you found a way to lose data and that this is just a patch for a larger issue. For example, does it also happen if I shut down my phone mid-migration, or switch to another app in some other way?

src/main/java/org/medicmobile/webapp/mobile/ClosableAppActivity.java
```diff
@@ -0,0 +1,77 @@
...
+			backToExitPressedOnce = true;
+
+			toastMessage();
+
+			new Handler().postDelayed(new Runnable() {
+				@Override public void run() {
+					// After WAIT_MILLIS clean the back pressed history
+					backToExitPressedOnce = false;
+				}
+			}, WAIT_MILLIS);
```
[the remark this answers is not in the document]
mrsarm: Yes, your implementation seems simpler, I just made in this way because I copy & pasted the solution from Stackoverflow :grin: 

> Why add the double back to exit feature? Is it just to block the default behaviour of the back button? I think it'd be cleaner to let a single back press behave normally unless migrating.

I've seen this behavior in Android when the app is in the middle of a process, because the user may react pressing back when an error is shown, but without a real intent to exit the app.

In the case of the login page, the user can enter the app again without losing any state or data, so going to remove the double press behavior for the login, and keep the rejection of leaving the app when the connection error happens while migrating.

src/main/java/org/medicmobile/webapp/mobile/Utils.java
mrsarm: Added the button in the activity, it shows the error description (though looks more like a code, and the error code number):

Medic Android - No Internet - More Info

We can create a ticket after merge this to make more appealing.

garethbowen: Awesome, thanks!

garethbowen (approved):
Great work! Thanks for the quick turnaround.

Just waiting on a green build and then it's ready for AT.

PR #175, opened 2021-05-10
[To be deprecated] 6981 RDToolkit integration

latin-panda:
**Description**

In this PR: 
- Add RDToolkit library that needs Jetifier and AndroidX.
- Adds RDToolkitSupport.java and RDToolkitSupportActivity.java to create intents that communicates with RDToolkit app.
- Adds unit test libraries and coverage to RDToolkitSupport and RDToolkitSupportActivity.
- Separates execution of static-check, lint and test tasks (test xwalk and test webview). 

Ticket: https://github.com/medic/cht-core/issues/6981
CHT-Core integration PR: https://github.com/medic/cht-core/pull/7093
CHT-Docs: https://github.com/medic/cht-docs/pull/510

**License**


latin-panda:
Hi @dianabarsan even though we can't merge this now, I'd love to get some feedback in this code, can you please have a look? Thanks!

src/main/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
```diff
@@ -0,0 +1,219 @@
...
+
+	private final Activity ctx;
+
+	RDToolkitSupport(Activity ctx) {
+		this.ctx = ctx;
+	}
+
+	String process(int requestCode, int resultCode, Intent intentData) {
+
+		log(this, "RDToolkitSupport :: process requestCode=%s", requestCode);
```
dianabarsan, line 44: Should this be debug? (this applies to the other messages that this file logs.

src/main/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
```diff
@@ -0,0 +1,219 @@
...
+		}
+
+		return intent;
+	}
+
+//> PRIVATE HELPERS
+
+	private String makeProvisionTestJavaScript(Object response) {
+		String javaScript = "try {" +
+				"const api = angular.element(document.body).injector().get('AndroidApi');" +
```
dianabarsan, line 122: Are we planning to integrate this into older versions of the app (before 3.10)? 
If not, we can use `window.CHTCore.AndroidApi` instead. The `angular.element...` is deprecated as on 3.11

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -231,6 +239,27 @@ protected void onStart() {
...
+						log(this, "RDToolkitSupport :: Executing JavaScript: %s", provisionScript);
+						evaluateJavascript(provisionScript);
+						return;
+
+					case RDTOOLKIT_CAPTURE_ACTIVITY_REQUEST_CODE:
+						int readPermission = ContextCompat.checkSelfPermission(this, READ_EXTERNAL_STORAGE);
+
+						if (readPermission != PERMISSION_GRANTED) {
+							log(this, "RDToolkitSupport :: Requesting storage permissions");
+							ActivityCompat.requestPermissions(this, PERMISSIONS_STORAGE, RDTOOLKIT_CAPTURE_ACTIVITY_REQUEST_CODE);
```
dianabarsan, line 254: This respects the "prominent disclosure" rules, right?

dianabarsan: Also, is this an async or a sync operation? Will the thread "halt" until the user has granted permission, or will the capture script that follows run anyway (and fail)?

dianabarsan (commented):
Nice, very clean! 
I left a couple of questions inline.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: > This respects the "prominent disclosure" rules, right?

This permission is already in use in medic-android, but as I read, this should be fine. They did put additional restrictions for MANAGE_EXTERNAL_STORAGE permission (which we still don't require) in Android 11+. 

> Also, is this an async or a sync operation? Will the thread "halt" until the user has granted permission, or will the capture script that follows run anyway (and fail)?

I'm using Android 11.
If the user remove the permission in the phone settings, then when capturing an RD Test result the phone displays a modal asking for granting the permission again.
if the user close the modal (not granting, not rejecting) the results aren't capture but user can continue using the app normally, it's not breaking the app. 
if the user reject the permission, the results aren't capture and the user can use the app normally, it's not breaking the app neither.

src/main/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
latin-panda: Because we're using Angular 10, I don't think we're going to support 3.10 and before. 
Switching to use: `window.CHTCore.AndroidApi` 💪

src/main/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
latin-panda: I've changed them by trace()

build.gradle
```diff
@@ -337,6 +337,8 @@ dependencies {
   implementation fileTree(dir: 'libs', include: ['*.jar'])
   implementation 'com.simprints:LibSimprints:1.0.11'
   implementation 'com.github.Mariovc:ImagePicker:1.2.2'
+  implementation 'com.github.dimagi:rd-toolkit:0.9.0'
```
latin-panda, line 340: Hi @mrsarm I tried moving the class to inside the webview folder but still getting the same error, I wonder if because the lib is introduced here and taken into account in all builds 🤔
Screen Shot 2021-05-13 at 3 57 38 pm

BTW I tried building for Android 9 but geting an error where sdk isn't reaching the minimum supported by our webview build:
Screen Shot 2021-05-13 at 3 57 10 pm

I think maybe we should make a flavour for this partner until we make webview fully supported across android versions, what do you think?

mrsarm: To make it compile you have to change the `implementation` call with `webviewImplementation` with the libraries added:

```diff
-   implementation 'com.github.dimagi:rd-toolkit:0.9.0'
-   implementation 'androidx.core:core-ktx:1.0.1'
+   webviewImplementation 'com.github.dimagi:rd-toolkit:0.9.0'
+   webviewImplementation 'androidx.core:core-ktx:1.0.1'
```

So they are only included in the Webview flavors, and remove the `android.enableJetifier=true` property added that is not compatible with XWalk. If that property is really needed, you should find a way to only add it when compiling Webview flavors.

mrsarm: :thinking: removing `android.enableJetifier=true` causes the project to compile but not to build in Webview flavors, so you have to figure out how to add that property only for Webview flavors, so far I couldn't find a simple way to do it, it is easy from the command line via arguments but must be a way to override the property regarding the flavor.

latin-panda: Thanks Mariano, that's good stuff!! 
Yes, we need the `android.enableJetifier=true` because the RDToolkit lib is using the new [android support lib](https://developer.android.com/jetpack/androidx) (backward compatible) and if we use the old android.support.v4 lib (not maintenance since 2018) throws an error of incompatibility.
Actually, since we're using both flags, then maybe I dont need to import the androidx manually (trying out...)

latin-panda: I tried to add this flag inside the webview flavor (build.gradle) but it's not working, the posts I've found say that these android flags must go in the gradle.properties 
At the end, I moved it to `deploy-flavour` command in makefile and the build works, but then the `make test` is failing. 
If I add it in the `make test` then it works for webview but not for xwalk. I'm trying to find where we are configuring the test.

latin-panda: This is fixed, I've put back the `android.enableJetifier=true` in gradle.properties to migrate third parties with the new android [support lib](https://developer.android.com/jetpack/androidx) in webview since that's the way in the future, and disable the flag in xwalk

latin-panda:
Hi guys, @mrsarm @dianabarsan 
Can you please take a look again? Thanks!

src/webview/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
```diff
@@ -0,0 +1,232 @@
...
+			throw new RuntimeException("RDToolkitSupport :: Bad result code for capturing result: " + resultCode);
+		}
+
+		try {
+			JSONObject response = parseCaptureResponseToJson(intentData);
+			return makeCaptureResponseJavaScript(response);
+
+		} catch (Exception exception) {
+			error(exception, "RDToolkitSupport :: Problem serialising the captured result");
+			return safeFormat("console.log('Problem serialising the captured result: %s')", exception);
```
dianabarsan, line 143: As above, if we log these as errors, is a feedback doc created?

src/webview/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
```diff
@@ -0,0 +1,232 @@
...
+			throw new RuntimeException("RDToolkitSupport :: Bad result code for the provisioned RD Test: " + resultCode);
+		}
+
+		try {
+			JSONObject response = parseProvisionTestResponseToJson(intentData);
+			return makeProvisionTestJavaScript(response);
+
+		} catch (Exception exception) {
+			error(exception, "RDToolkitSupport :: Problem serialising the provisioned RD Test");
+			return safeFormat("console.log('Problem serialising the provisioned RD Test: %s')", exception);
```
dianabarsan, line 158: Same here, does an error create a feedback document?

src/webview/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -365,7 +385,7 @@ private void logException(Exception ex) {
 				.replace("\n", "; ")
 				.replace("\t", " ");
 
-		parent.errorToJsConsole("Execption thrown in JavascriptInterface function: %s", stacktrace);
+		parent.errorToJsConsole("Exception thrown in JavascriptInterface function: %s", stacktrace);
```
dianabarsan, line 388: Nice! I cannot tell you for how long this typo has bugged me! (ranted about it here: https://github.com/medic/cht-core/issues/6162)

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -389,10 +431,16 @@ public boolean getLocationPermissions() {
...
 	public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
 		super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-		if (requestCode != ACCESS_FINE_LOCATION_PERMISSION_REQUEST_CODE) {
+
+		if (requestCode == ACCESS_FINE_LOCATION_PERMISSION_REQUEST_CODE) {
+			locationRequestResolved();
+			return;
+		}
+
+		if (requestCode == ACCESS_STORAGE_PERMISSION_REQUEST_CODE) {
```
dianabarsan, line 440: We're only adding this for the debug message?

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -300,6 +312,36 @@ public void errorToJsConsole(String message, Object... extras) {
...
+		evaluateJavascript(provisionScript);
+	}
+
+	private void processRdToolkitCaptureResultActivity(int requestCode, int resultCode, Intent intentData) {
+		int readPermission = ContextCompat.checkSelfPermission(this, READ_EXTERNAL_STORAGE);
+
+		if (readPermission != PERMISSION_GRANTED) {
+			trace(this, "RDToolkitSupport :: Requesting storage permissions");
+			ActivityCompat.requestPermissions(this, PERMISSIONS_STORAGE, ACCESS_STORAGE_PERMISSION_REQUEST_CODE);
+			return;
```
dianabarsan, line 337: So, we return here if permission isn't granted and we request it. Does this mean that the capture script that follows never runs now? Or is this code somehow triggered again? 

If not, could we extract the capture script (the rest of this function) in a separate function and maybe also run it when `ACCESS_STORAGE_PERMISSION_REQUEST_CODE` is received?

dianabarsan (commented):
Nice work! I left a couple of comments and questions inline.

mrsarm (changes requested):
Sorry @latin-panda I couldn't review it properly today, although @dianabarsan covered that well !

One comment I can add is that I noticed that you left that flag that is needed in Webview but makes XWalk to fail, and although you manage the change of the flag in the Makefile and the build scripts, the change makes the build to fail when compiling the project with Android Studio to output one of the XWalk flavors, same if Gradle is used in the command line to also build a specific flavor with XWalk.

I understand that Gradle doesn't provide a more flexible way to change the flag regarding the flavor, so could you please add a note in the README file, under the development section? so developers don't get crazy specially when trying to debug the XWalk version with Android Studio.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: The idea of user experience is:
- The user accepts the permission
- then click in "capture" button. That's why no action [here](https://github.com/medic/medic-android/pull/175/files/6776128ea7d3a4a59102c0837d9d96a704f851c8#r637007069) 
- medic-android closes 
- RDtoolkit app opens and request the user to capture results

When testing I didn't feel it was a weird user experience and I feel it gives a transition experience when switching apps but I'm the only one using this so far and not the target user so feedback on this very valuable. 

What do you think? Better to trigger the action automatically after the permission is granted? The experience would be: 
- The user accept the permission
- Immediately medic-android closes
- RDToolkit app opens and request the user to capture the results.

src/webview/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
latin-panda: Done!

src/webview/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
latin-panda: ✨

src/webview/java/org/medicmobile/webapp/mobile/RDToolkitSupport.java
latin-panda: Done, yes it does

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
dianabarsan: But the permission isn't requested every time, right? If the permission is already granted, we're already in the 2nd situation.

dianabarsan: I'm also unsure what triggers the "RDTOOLKIT_CAPTURE_ACTIVITY_REQUEST_CODE". Is it the user clicking on the "capture" button?

latin-panda: > But the permission isn't requested every time, right? 

If the app already has the permission granted then it wont ask for it again. The experience is when clicking the capture button then the RDToolkit app opens (no modal asking for permission before opening RDToolkit) 

> I'm also unsure what triggers the "RDTOOLKIT_CAPTURE_ACTIVITY_REQUEST_CODE". Is it the user clicking on the "capture" button in the first place? If there's a permission request, then the user needs to click capture twice?

Yes, if the permission isn't granted, then click capture button, modal appears, accept permission, and click capture button, RDToolkit open. 

We reach this android code when clicking on the button, I can try checking the permission somewhere else (tEmbeddedBrowserActivity > onStart() maybe ?), probably the modal asking for permissions will display when opening the medic-android (I guess ?)  
Maybe this is better experience and will cover any other feature needing the same permission 🤔

dianabarsan: I think we should be consistent about how the app behaves with or without the permission. It's not intuitive that you need to click the button the 2nd time if you didn't have permission.

dianabarsan: So we should trigger the "capture" part when the permission is granted, like in my initial comment maybe?

Makefile
```diff
@@ -9,33 +9,51 @@ ifdef ComSpec	 # Windows
...
-	${GRADLEW} androidCheck lintUnbrandedWebviewDebug test
+
+lint-webview:
+	${GRADLEW} androidCheck lintUnbrandedWebviewDebug
+
+test-webview:
+	${GRADLEW} -Pxwalk=false test
+
+test-xwalk:
+	${GRADLEW} -Pxwalk=true -Pandroid.enableJetifier=false test
```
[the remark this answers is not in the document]
latin-panda: androidCheck runs for both.. I think we're going to replace it because jcenter is dead
For now I've moved the androidCheck to it's own task in makefile: `static-check`

src/testWebview/java/org/medicmobile/webapp/mobile/RDToolkitSupportTest.java
```diff
@@ -0,0 +1,342 @@
...
+		when(RdtUtils.getRdtSession(intent)).thenReturn(session);
+
+		JSONObject expectedResponse = json(
+				"sessionId", "session1",
+				"timeResolved", "2020-05-25T22:20:05.000+0000",
+				"timeStarted", "2020-05-25T22:15:00.000+0000",
+				"state", "RUNNING"
+		);
+
+		String expectedScript = "try {" +
```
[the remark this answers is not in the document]
latin-panda: Yeah, I researched before but couldn't find a way yet, also we're not returning a value so it's like checking it's not throwing uncontrolled exceptions,  It'd be nice running lint in those scripts 🤔

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: Fixed

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: Fixed

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -389,10 +420,22 @@ public boolean getLocationPermissions() {
 	@Override
 	public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
 		super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-		if (requestCode != ACCESS_FINE_LOCATION_PERMISSION_REQUEST_CODE) {
+		boolean granted = grantResults.length > 0 && grantResults[0] == PERMISSION_GRANTED;
+
+		if (requestCode == ACCESS_FINE_LOCATION_PERMISSION_REQUEST_CODE) {
```
latin-panda, line 425: The recommended practice is to check if the permission is granted or denied ([article](https://developer.android.com/training/permissions/requesting)), otherwise is like not considering the user decision (as tested in Samsung Android 11). 
We'll be handling this for Location + some refactoring in this Activity in another ticket because of the location testing.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -389,10 +420,22 @@ public boolean getLocationPermissions() {
...
+			locationRequestResolved();
+			return;
+		}
+
+		if (requestCode == ACCESS_STORAGE_PERMISSION_REQUEST_CODE) {
+			if (granted) {
+				this.rdToolkitSupportActivity.resumeCaptureActivity();
```
dianabarsan, line 432: NIIIIICE!

dianabarsan (approved):
Great work! Love it!

mrsarm (approved):
Great work ! LGTM :+1:

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: Ticket: https://github.com/medic/cht-android/issues/189

mrsarm:
Hey @latin-panda , hold down the PR a bit, I'm starting to move the codebase to Gradle 7 and Java 8, and looks like Gradle 7 allows to have different `gradle.properties` files for each flavor, and although you manage well the situation of the XWalk flavors not supporting some keys in that file, we still need to override the value in Android Studio so using Gradle 7 would make the switching easier. Will confirm later.

latin-panda:
Deprecated by a different approach, please refer to related issue for more details.

PR #179, opened 2021-05-25
Switch to cht-android

dianabarsan:
medic/cht-core#6586


dianabarsan:
I've tested `make`-ing this app on top of an existent installation of an older version of `medic-android` and: 
image

Changed to: 
image

Settings look like: 
image

dianabarsan:
Please review, @mrsarm . Thanks in advance!

settings.gradle
```diff
@@ -1 +1 @@
-rootProject.name = 'medic-android'
\ No newline at end of file
+rootProject.name = 'medic-android'
```
mrsarm, line 1: cht-android ?

mrsarm (changes requested):
I think you forgot to change the name in `settings.gradle`. After change the value there, the name of the APKs will also change, so we need to update these references too:

1. [Fastfile#L22-L26](https://github.com/medic/medic-android/blob/3c62a789bf01369410c5086a538e1baec9a3660b/fastlane/Fastfile#L22-L26)
2. [README.md#L13](https://github.com/medic/medic-android/blame/3c62a789bf01369410c5086a538e1baec9a3660b/README.md#L13)
3. [README.md#L58](https://github.com/medic/medic-android/blame/3c62a789bf01369410c5086a538e1baec9a3660b/README.md#L58)

settings.gradle
dianabarsan: Yea, I don't know if we want to change those too. If we want the APK names to change and everything.

dianabarsan: I looked into this as well, but in the end, wanted to play it safe.

mrsarm: Oh I see you added the question [here](https://github.com/medic/cht-core/issues/6586#issuecomment-848877731) whether to rename the APKs or not. I would say yes but lets wait Gareth's answer.

mrsarm (approved):
LGTM  :100:

PR #182, opened 2021-06-08
6741 - Adding device info to about page

latin-panda:
# Description

Adding more info to getDeviceInfo method that will be used in about page: packageName and versionCode.

Telemetry is already saving this data [here](https://github.com/medic/cht-core/blob/28f2afb6bf5870ba73aabf8f8f20b237a05648ab/webapp/src/ts/services/telemetry.service.ts#L161).

CHT-Core: https://github.com/medic/cht-core/pull/7131
CHT-Docs: https://github.com/medic/cht-docs/pull/514

https://github.com/medic/cht-core/issues/6741

# License


mrsarm (approved):
Nice, LGTM!

PR #185, opened 2021-06-08
feat: build with itech aurum

freefony:


src/itech_aurum/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    <background android:drawable="@color/ic_launcher_background"/>
+    <foreground android:drawable="@mipmap/ic_launcher_foreground"/>
+</adaptive-icon>
```
mrsarm, line 5: Please add a return line character at the end of the file.

src/itech_aurum/res/mipmap-anydpi-v26/ic_launcher_round.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    <background android:drawable="@color/ic_launcher_background"/>
+    <foreground android:drawable="@mipmap/ic_launcher_foreground"/>
+</adaptive-icon>
```
mrsarm, line 5: Same about EOL.

mrsarm (changes requested):
Looks mostly good, but added some suggestions inline, and also:

1. You need to add a entry in the `publish.yml` file like [this](https://github.com/medic/cht-android/blob/4a306a982c60cbc01f341373bfe31114724bd97a/.github/workflows/publish.yml#L49-L53), so when a new version is tagged , Fastlane will yield the APK for your brand.
2. Consider adding a manifest exactly like [this](https://github.com/medic/cht-android/blob/4a306a982c60cbc01f341373bfe31114724bd97a/src/bracuganda/AndroidManifest.xml), where the file permission is removed. The permission cannot be used anyway because file uploads is not currently working and until the next majar version (4.0.0) we don't plan to fix it. Leaving the perm won't hurt the app but may add friction in the approval if you are going to publish the app in Google Play, causing the app to even be rejected.

mrsarm (approved):
LGTM :+1:

PR #187, opened 2021-06-10
Add itech malawi

freefony:


src/itech_malawi/res/values/ic_launcher_background.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="ic_launcher_background">#FFFFFF</color>
+</resources>
```
mrsarm, line 4: Add end of line please.

mrsarm (changes requested):
Hey @freefony , I think you forgot to add the `strings.xml` file with the name and URL to the CHT app, right? Like the one you added in your last PR [here](https://github.com/medic/cht-android/pull/185/files#diff-6d0116f33263488f47ea22227e6fd07a299fc7016b70ca77663b8ff9fd3cbfad).

mrsarm (approved):
LGTM :+1:

PR #188, opened 2021-06-15
Itech malawi/change app name

freefony:


src/itech_malawi/res/values/strings.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="app_name">Itech Malawi</string>
+    <string name="app_name">2WT Retention App</string>
     <string name="app_host">itech-malawi.app.medicmobile.org</string>
 </resources>
```
freefony, line 5: is the build somehow removing the EOL? i am very sure i added it
Screen Shot 2021-06-15 at 1 36 25 PM

mrsarm: I had the same impression in the past, I think it may be the editor which does the removal, not git or the build tools.

Anyway, just in case check your `~/.gitconfig` file, there are configurations for that but I don't remember which ones.

mrsarm (approved):

Issue #191, opened 2021-07-01
CI fails because something in the AVD tests changed the locale to French


mrsarm:
All CI pipelines for this project started to fail yesterday because something related to the AVD instance used to run some of the tests:

```
> Task :connectedUnbrandedWebviewDebugAndroidTest
Starting 2 tests on test(AVD) - 10

org.medicmobile.webapp.mobile.SettingsDialogActivityTest > testLoginScreen[test(AVD) - 10] FAILED 
	androidx.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError: 'a string containing "Incorrect user name or password. Please try again."' doesn't match: Nom d'utilisateur ou mot de passe incorrect. Veuillez réessayer
	Expected: a string containing "Incorrect user name or password. Please try again."
```

Looks like something related with the localization, the tests expect the instance to run an AVD with English configuration by default, but French is used so the text are localized with that language.

Maybe setting the language in the login page to English before doing anything else can solve the problem.

Issue #174, opened 2021-05-07
Update gradle version and relevant dependencies 


ngaruko:
As we move on supporting later android sdks (29, 30), some features are becoming hard to support without updating the gradle and some dependencies.

mrsarm:
Hey @ngaruko , could you detail what are the cases where the Gradle version is giving you errors or deprecation warnings?

I want also as you say determine whether we need a newer version of Gradle for the SDK upgrade (https://github.com/medic/medic-android/issues/164). At a first glance, we do need to upgrade in order to use the latest Android plugin, checkout https://developer.android.com/studio/releases/gradle-plugin#updating-gradle , but not sure if we need one of the latest versions of the plugins in order to upgrade the Android SDK.

Also worth to note:

- JCenter is one of the main repos used by Gradle and our project to get the dependencies, and it's going to shutdown in a few months, and the replacement for some of the dependencies is a repo for Google, at least for the Android dependencies, check https://developer.android.com/studio/releases/gradle-plugin#updating-plugin
- We are not supporting Java syntax newer than Java 7, which though is not a big problem but annoying not being able to use important changes introduced in Java 8, may be a problem in the future. Now we can use newer compilers and Gradle takes care of setting the right target version, but at some point that may not work well, specially because using higher versions of compiler to target older versions can lead to errors like using core API libraries that are not available in the older versions and not detected by the compiler.

latin-panda:
++ 1

I'm also receiving the warning regarding JCenter and Gradle, also definitely would love some features of Java 8.

Mockito has a lib for Android with tools for testing static and final implementations which can help us getting coverage to places where we don't have control, but some of these tools need Java8+.

I think we might need to add [this](https://developer.android.com/studio/write/java8-support#library-desugaring) for Java 8:

``` java
android {
    defaultConfig {
        // Required when setting minSdkVersion to 20 or lower
        multiDexEnabled true
    }

    compileOptions {
        // Flag to enable support for the new language APIs
        coreLibraryDesugaringEnabled true
        // Sets Java compatibility to Java 8
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
}
```

ngaruko:
[...]

Issue #164, opened 2021-03-17
Upgrade target sdk to 30 and associated changes


garethbowen:
Update `targetSdkVersion` to 30 along with a range of other requirements, [documented here](https://developer.android.com/distribute/best-practices/develop/target-sdk).

New apps will require this by August 2021
App updates will require this by November 2021

garethbowen:
With the change to the app bundle format we'll need to make sure the app can still be sideloaded, and ideally still able to be distributed via F-Droid.

mrsarm:
### Progress Update

- [x] The requirement for new and existing Apps to upgrade the targeted **SDK to version 30** (Android 11) was [implemented](https://github.com/medic/cht-android/pull/194) and merged into master this week.
- [x] **We already complaint with other new requirements for existing apps**, e.g. the minimal version of the algorithm to sign APKs and Bundles is v2, that is the version we use now.
- [x] The are other requirements that we have verified don't apply to our apps, e.g. location permission in the background has new requirements but we only access location in the foreground.
- [ ] New apps in the Play Store (not updates from existing apps) will require to be uploaded using the new Bundle format (.aab). We did some experiments with this format and creates this files is an easy task. The bundle cannot be used for side-loading, but we can still generate the .apk files for each flavor along with the .aab files. We will implement a specific task in the CI to produce the .aab files for new flavors.
- [ ] New apps in the Play Store (not updates from existing apps) after being uploaded with the new Bundle format will be converted by Google to optimized APKs in the Play Store, so Google need the same or a new created private key to sign the app. We are defining a new workflow for this.

#### Proposed workflow

In summary, the new  [workflow](https://docs.google.com/document/d/1ZAiY-67DHdX7AvFxZoXo_FLNwztG6zYdeN9PcwxFtmk/edit?usp=sharing) is:

- New apps from partners will have their own private keys to sign the app, not using the one that we currently use to sign existing apps.
- We will use the same set of keys to sign the bundle (Upload Key) and to let Google sign the APKs (Signing App Key), allowing us to keep generating .apk files for sideloading compatible with the .apk files generated by Google (signed with the same key).
- We will make changes in the CI pipelines to generate both .apk and .aab files, signed with this key, so each time a new private key is generated to sign an app, we need to setup it in our repo. We need to document well each step to create a new flavored version of the app. We also recommend to only setup one key for partner, using the new key for each new flavored app created by that partner.
[...]

PR #192, opened 2021-07-02
Upgrade Java dependencies and fix instrumentation test

mrsarm:
Upgrade building dependencies (https://github.com/medic/cht-android/issues/174), and fix broken CI because gamma dev environment started to use French as default language (https://github.com/medic/cht-android/issues/191).

- Upgrade dependencies (https://github.com/medic/cht-android/issues/174):
  - Min Java ~6~ → 8
  - Gradle ~6.5~ → 7.1
  - Android Gradle Plugin ~4.1.3~ → 4.2.2
  - Other minor upgrades of building and testing libraries
-  Fix login test broken when development environment use French as default language instead of English (https://github.com/medic/cht-android/issues/191).
- Fix lint configuration that now works again when used with the XWalk builds.
- Refactor some code replacing old anonymous Java classes with lambda style (Java 8
- Improvements in the _Makefile_ scripts, add ability to set the flavor in some of the targets. 
- Fix indentation in `build.gradle`.
- Add more info in the _"Development"_ section in the README.


README.md
```diff
@@ -12,12 +12,13 @@ To help you pick which APK to install you can find information about the version
 
 The APKs are named as follows: `cht-android-{version}-{brand}-{rendering-engine}-{instruction-set}-release.apk`
 
-| Rendering engine | Instruction set | Android version | Notes |
-|------------------|-----------------|-----------------|--|
-| `webview`        | `arm64-v8a`     | 10+             |  Preferred. Use this APK if possible. |
+| Rendering engine | Instruction set | Android version | Notes                                                       |
+|------------------|-----------------|-----------------|-------------------------------------------------------------|
+| `webview`        | `arm64-v8a`     | 10+             | Preferred. Use this APK if possible.                        |
 | `webview`        | `armeabi-v7a`   | 10+             | Built but not compatible with any devices. Ignore this APK. |
```
latin-panda, line 18: Unrelated question to this PR:

Why do we build the `armeabi-v7a` APK if it can't be used? 🤔 Can we ask to remove it from the build to avoid confusion?

README.md
```diff
@@ -91,22 +92,39 @@ This release changes the way in which location data is collected to better align
...
 - [bug] [cht-core#6648](https://github.com/medic/cht-core/issues/6648): Blank screen when launching external apps from CHT Android app
 
+
 # Development
 
-1. Install Android SDK
-2. Clone the repo
-3. Plug in your phone. Check it's detected with `adb devices`
-4. Execute: `make` (will also push app unto phone)
+1. Install the [Android SDK](https://developer.android.com/studio#command-tools) and Java 8+ (it works with the OpenJDK versions).
```
latin-panda, line 98: Do we need to specify that we're using gradle 7?

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/SettingsDialogActivityTest.java
```diff
@@ -121,7 +123,7 @@ public void testLoginScreen() throws Exception {
 				.perform(webClick());
 		Thread.sleep(4000);//TODO: use better ways to handle delays - takes longer with emulators
 		onWebView().withElement(findElement(Locator.CSS_SELECTOR, "p.error.incorrect"))
-				.check(webMatches(getText(), containsString(ERROR_INCORRECT)));
+				.check(webMatches(getText(), matchesPattern(ERROR_INCORRECT)));
```
latin-panda, line 126: Is it possible to select the language in the login page instead? 
Because if we decide to add more gamma testing, it's going to be difficult to maintain 2 labels every time.

src/main/java/org/medicmobile/webapp/mobile/AlertDialogUtils.java
```diff
@@ -31,11 +30,7 @@ public static void show(
 		new AlertDialog.Builder(context)
 			.setTitle(title)
 			.setMessage(message)
-			.setNeutralButton(okLabel, new DialogInterface.OnClickListener() {
-				@Override public void onClick(DialogInterface dialog, int which) {
-					dialog.dismiss();
-				}
-			})
+			.setNeutralButton(okLabel, (dialog, which) -> dialog.dismiss())
```
latin-panda, line 33: ✨ So nice to do this! ✨

src/xwalk/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -253,19 +253,17 @@ SmsSender getSmsSender() {
...
-					container.evaluateJavascript(js, IGNORE_RESULT);
-				}
+		container.post(() -> {
+			// `WebView.loadUrl()` seems to be significantly faster than
+			// `WebView.evaluateJavascript()` on Tecno Y4.  We may find
+			// confusing behaviour on Android 4.4+ when using `loadUrl()`
+			// to run JS, in which case we should switch to the second
+			// block.
+			// On switching to XWalkView, we assume the same applies.
+			if(true) { // NOPMD
```
latin-panda, line 263: I know this code was there before.... but if it's always true, can we remove the `else` section ?
Do you think this big comment can be placed before the `container.post`? then we can have 1 line function 🤔

latin-panda (changes requested):
Nice Job ✨ I'm so happy to have this upgrade, it'll provide a lot of flexibility!!
I've left some comments and questions in line, nothing big.
Also, I've done a quick smoke test in Android 11 and 9 and look okay, but I'd recommend to test in older devices, I think Bede has an Android 6.

README.md
garethbowen: I think it automatically builds every combination of every dimension, so I don't think we can stop it. We _could_ delete it when the build finishes with a custom script. It's probably not worth spending time on at this point as it will almost certainly go away when we solve [this issue](https://github.com/medic/cht-android/issues/164) because after that we expect to build one bundle instead of a series of apks.

README.md
mrsarm: No because the `gradlew` script takes care of that, downloading Gradle 7.1 in the user space the first time is executed, and the build is still compatible with older versions of Gradle, but I'll add a note about it.

src/xwalk/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Yes didn't notice the if inside when refactoring.

README.md
```diff
@@ -91,22 +92,39 @@ This release changes the way in which location data is collected to better align
...
+To execute unit tests: `./gradlew testUnbrandedWebviewDebugUnitTest`.
 
-# Testing - Instrumentation Tests (UI Tests)
+### Instrumentation Tests (UI Tests)
 
-1. Install Android SDK
-2. Clone the repo
-3. Plug in your phone. Check it's detected with `adb devices`
-4. Execute: `./gradlew connected[Flavor]WebviewDebugAndroidTest` . Eg `./gradlew connectedUnbrandedWebviewDebugAndroidTest` or `./gradlew connectedMedicmobilegammaWebviewDebugAndroidTest`. At the moment we have tests only in these 2 flavors: unbranded and medicmobilegamma. To avoid failures running the tests, previous versions of the app should be uninstalled first, otherwise an `InstallException: INSTALL_FAILED_VERSION_DOWNGRADE` can make the tests to fail, and Android needs to have English as default language.
+These tests run on your device.
```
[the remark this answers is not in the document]
mrsarm: Yes we only support physical devices from Gradle and Make to run and test the app.

mrsarm:
@latin-panda please review again, here are the changes since your last review:

- Refactor and improvements of the Makefile, to support the flavor argument in some of the targets, and move boilerplate configuration to env variable.
- Moved some code that was only used by the Webview version that caused the linter to fail when executing the check for the Xwalk version (`make flavour=UnbrandedXWalk lint`).
- Changed the documentation using `make` command instead of the Gradle versions


Also I forget to mention that one of the goal of this PR was to remove the **JCenter** repository, but we cannot do that due that we have two dependencies that are only hosted in JCenter, but according with [this](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/), JFrog, the repo's owner, quote:

> _...will keep JCenter as a read-only repository indefinitely. Our customers and the community can continue to rely on JCenter as a reliable mirror for Java packages._

latin-panda (commented):
Nice job!! ✨ I think the makefile commands are super easy now, thanks!!
I left a couple of suggestion in the documentation, the rest is fine!

README.md
```diff
@@ -91,22 +92,54 @@ This release changes the way in which location data is collected to better align
...
+3. Plug in your phone. Check it's detected with `adb devices`.
+4. Execute: `make` (will also push app into the phone). Use `make xwalk` for the Xwalk version instead.
+
+Gradle is also used but it's downloaded and installed in the user space the first time `make` is executed.
+
+You can also build and launch the app with [Android Studio](#android-studio).
+
+## Flavor selection
+
+Some *Make* targets support the flavor and rendering engine as `make flavour=[Flavor][Engine] TASK`, where `[Flavor]` is the branded version with the first latter capitalized and the engine is either `Webview` or `Xwalk`. The default value for `flavor` is `UnbrandedWebview`, e.g. executing `make deploy` will assemble and install that flavor, while executing `make flavour=MedicmobilegammaXwalk deploy` will do the same for the _Medicmobilegamma_ brand and the `Xwalk` engine.
```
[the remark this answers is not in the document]
latin-panda: I'm wondering if the `make` argument `flavour=` should be `flavor`... looking at the android docs, it appears that they use `flavor` 🤔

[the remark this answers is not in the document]
mrsarm: Yes I guess is an old typo, fixing...

README.md
```diff
@@ -91,22 +92,39 @@ This release changes the way in which location data is collected to better align
...
+3. Plug in your phone. Check it's detected with `adb devices`.
+4. Execute: `make` (will also push app into phone).
+
+To only **build and assemble** the APKs for certain flavor and rendering engine, execute `./gradlew assemble[Flavor][Engine]Debug`, e.g. to assemble the unbranded flavor for the webview rendering engine: `./gradlew assembleUnbrandedWebviewDebug`. For the XWalk engine: `./gradlew assembleUnbrandedXWalkDebug`.
+
+The APKs are stored in the folder `build/outputs/apk/[Flavor][Engine]/debug/`, in the example above: `build/outputs/apk/unbrandedWebview/debug/`.
+
+To only execute the **linter checks**, execute: `./gradlew androidCheckstyle`.
+
+The [Android Studio](https://developer.android.com/studio) can be used to build and launch the app instead. Be sure to select the right flavor and rendering engine from the _Build Variants_ dialog (see [Build and run your app](https://developer.android.com/studio/run)). To launch the app in an emulator, you need to uncomment the code that has the strings for the `x86` or the `x86_64` architecture in the `build.gradle` file, in the `android` / `splits` / `include` section.
```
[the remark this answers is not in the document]
mrsarm: I made it a sub section, but the links provided should be enough for the user to go deeper of how to use Android Studio and how to setup emulators. I only wanted to mention about the selection of the variant and the commented emulators' arch because are the typical question that I had and others had in the past when starting to work with the app.

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/SettingsDialogActivityTest.java
```diff
@@ -92,7 +92,7 @@ public void testLoginScreen() throws Exception {
 								0)))
 				.atPosition(2);
 		linearLayout.perform(click());
-		Thread.sleep(10000);//TODO: use better ways to handle delays
+		Thread.sleep(12000);	//TODO: use better ways to handle delays
```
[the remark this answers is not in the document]
mrsarm: `Thread.sleep(..)` is a really bad option when running multiple threads, but here we only run once at a time, so for now I'll keep it like that.

latin-panda (approved):
LGTM! This PR includes great improvements thanks!! ☀️🎉

mrsarm:
Going to merge without moving it to AT because only build and testing dependencies were updated (anyway I did many test locally with both XWalk and Webview flavors).

PR #193, opened 2021-07-12
164 - Meet new Google Play's target API level (SDK 30) and packaging requirements

mrsarm:
Upgrade of `targetSdkVersion` to 30 and added build instructions to build `.aab` files, along with a range of other Google Play new requirements.

Issue: #164

## Done
- [x] New targeted SDK ~29~ → 30 (Android 11).
- [x] Assemble AABs files for publishing in Google Play Store.

## TO-DO
- [ ] Check locales configuration when distributing the app with `.aab` format.
- [x] Release a beta version to check CI still works with the changes and releases the AAB files.
- [x] Check lane `deploy` although we are not going to enable it again for publishing in the Play Store.
- [x] Check what happens if we continue signing the new `.aab` files and then Google generates the `.apk` files from there, does Google in some way re-use the sign in the new files? if Google re-sign the APKs with its own signing keys, does it makes the upgrade process for devices with the app already installed to fail?

## No required

These are requirements mention in the Google's documentation (https://developer.android.com/distribute/best-practices/develop/target-sdk) that I have verified we already complain or don't affect our app.

- [x] [APK Signature Scheme v2 now required](https://developer.android.com/about/versions/11/behavior-changes-11#minimum-signature-scheme). We are already signing with v2 scheme. Steps used in the verification:
   1. Download one the last version of the App [Releases](https://github.com/medic/cht-android/releases), any of the APKs.
   2. Check the signature with `apksigner`:
   ```
   $ apksigner verify -v --print-certs cht-android-v0.8.0-5-unbranded-webview-arm64-v8a-release.apk
   Verifies
   Verified using v1 scheme (JAR signing): false
   Verified using v2 scheme (APK Signature Scheme v2): true
   Verified using v3 scheme (APK Signature Scheme v3): false
   Verified using v4 scheme (APK Signature Scheme v4): false
   Verified for SourceStamp: false
   Number of signers: 1
   Signer #1 certificate DN: CN=Unknown, OU=Unknown, O=MedicMobile, L=San Francisco, ST=California, C=CA
   Signer #1 certificate SHA-256 digest: 46e041d3de73ca6ed5844aac3e93728537b67a4e49449859b3c85cdfe41dde90
   ...
   ```
- [x] Do we need to change how we sign bundles letting Google doing that for us? → **No**. _(TODO add sources)_.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,7 @@ jobs:
         options: '{ "flavor": "unbranded" }'
 
   instrumentation-tests:
+    name: Instrumentation tests
     runs-on: macos-latest
     steps:
     - name: checkout
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,16 @@
 name: Build and test
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - 'README.md'
+    tags-ignore:
+      - *
+  pull_request:
+    paths-ignore:
+      - 'README.md'
+    tags-ignore:
+      - *
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
     paths-ignore:
       - 'README.md'
     tags-ignore:
-      - *
+      - v*.*.*
   pull_request:
     paths-ignore:
       - 'README.md'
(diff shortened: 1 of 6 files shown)


mrsarm:
Closing this PR because part was moved to another PR already merged: https://github.com/medic/cht-android/pull/194

And the rest (.aab files and signing changes) will be implemented separately.

PR #194, opened 2021-07-15
164 - Target Android 11 and other build changes

mrsarm:
Spin-off of the PR #193 , I move the change to target Android 11 that will be required next month for releasing new versions of existing apps, while leaving the original PR for the implementation of the new workflow for publishing new apps that will require more changes in the building and signing process.

### Changes
- New targeted SDK ~29~ → 30 (Android 11).
- Skip upload of APKs for webview-armeabi-v7a arch that is an architecture not used in devices with Android 10+.
- Skip CI build process if only README changes are pushed.
-  Allow `abi` argument to be overwritten in UI tests.
- `make assemble-all` only assemble release versions: there is a task to assemble the debug version (`assemble-all-debug`) and it's unlikely somebody will want to generate both. Furthermore, triggering the assembling of all versions cause the process to spend more than 2 Gb of space, and some time goes out of memory in doing so, even disabling the `--parallel` argument.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,25 +3,31 @@ name: Build and test
 on:
   push:
     paths-ignore:
-      - 'README.md'
+      - '**.md'
   pull_request:
     paths-ignore:
-      - 'README.md'
+      - '**.md'
 
 jobs:
+
   build:
+
     name: Build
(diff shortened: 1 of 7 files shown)


garethbowen (approved):
Awesome work. Good ideal splitting this change off from the bundle change!

PR #209, opened 2021-09-03
Upgrade Android Gradle Plugin → 7.0.2

mrsarm:
Upgrade to the latest Gradle plugin in order to make the linter to work again. Looks like we are going to have a PR like this each week :man_shrugging: .

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
     jcenter()
   }
   dependencies {
-    classpath 'com.android.tools.build:gradle:7.0.1'
+    classpath 'com.android.tools.build:gradle:7.0.2'
     classpath 'com.noveogroup.android:check:1.2.5'
   }
 }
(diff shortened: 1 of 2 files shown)


jkuester (approved):
Is there a particular reason the linter is forcing us to be on the latest patch of gradle?  Is this something we could configure in the linter so that we don't have to continuously bump this version (but instead can just update it in the same way we do the rest of the dependencies...)?

mrsarm:
> Is there a particular reason the linter is forcing us to be on the latest patch of gradle? Is this something we could configure in the linter so that we don't have to continuously bump this version (but instead can just update it in the same way we do the rest of the dependencies...)?

Right, it's a new check and looking into the documentation, we can disable it like other linters in [build.gradle](https://github.com/medic/cht-android/blob/85af8710941c6bce89156d0c302294c6035536f8/build.gradle#L154) with:

```groovy
disable 'AndroidGradlePluginVersion'
```

But to keep it performing the check I found a new trick that will keep it checking the version but not stopping the process with an error just for this: https://github.com/medic/cht-android/pull/209/files#diff-eefab2907910dd29e3d63b35664b4bdee63bb6b8b64d1238d8bee937f015fe9bR7.

PR #210, opened 2021-09-09
Remember previous URL when reloading app from Gopinath/medic changes

mrsarm:


garethbowen (changes requested):
The code looks good to me, thanks!

I know we don't do a lot of testing currently, but if possible can you put a little effort into writing a test for at least the webview version?

mrsarm:
> I know we don't do a lot of testing currently, but if possible can you put a little effort into writing a test for at least the webview version?

Hey @garethbowen , after trying different approaches to test this an other functionalities without instrumentation tests (e2e), I gave up and added a new e2e one. Mocking Android components in unit tests is really hard in webview apps.

One of the problems with e2e tests is that we have to load the webapp (the CHT webapp in our case). We have a few tests like that: [LoginTests.java](https://github.com/medic/cht-android/blob/master/src/androidTestMedicmobilegammaWebviewDebug/java/org/medicmobile/webapp/mobile/LoginTests.javal) and [SettingsDialogActivityTest.java](https://github.com/medic/cht-android/blob/master/src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/SettingsDialogActivityTest.java), but they only test basic stuff up to the login page using the gamma dev environment.

Features that are used beyond the login page requires to login, and we don't have stable dev environments with users and passwords that are fixed in those environments consistently, because the DBs are constantly rotated, and used by QA and developers to do manual testing , so we avoided for now tests beyond the login page.

Anyway because what we want to test in the app is the app code and not the CHT, for this particular test I have loaded another "HTML5" app instead of the CHT app: https://angular.io/resources?category=development , that is a single page app with links that change the content of a section without reloading the app but changing the URL in the browser, as the CHT webapp does when clicking any of the tabs. Here is a recording of the new automated test:

https://user-images.githubusercontent.com/1608415/134549325-e9805b87-468b-4d4b-90c6-fed28a01a35d.mp4

I have to change the page used in the test, and I was thinking to use something static that we have for components, and I found this https://docs.communityhealthtoolkit.org/atp/ . The components and [source code](https://github.com/medic/atp) look outdated, but I can work in a PR to add a new page with the behavior we need to test for this, and more broadly to use later in new tests, e.g. to test the location request workflow using this [page](https://docs.communityhealthtoolkit.org/atp/location/) in the same project.

What you think? I wanted your opinion before move forward.

garethbowen:
Nice work on the test!

The ATP solution is good enough for now, because it's less likely to randomly change and break this test.

I think eventually we will have to spawn a cht-core instance so we can test cht-android thoroughly but we can leave this as an exercise for another day. Great work!

mrsarm:
> Nice work on the test!
> 
> The ATP solution is good enough for now, because it's less likely to randomly change and break this test.
> ...

Thanks @garethbowen , in that case, would you mind to approve it so I ask the QA team to review it before merging?

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/LastUrlTest.java
```diff
@@ -0,0 +1,79 @@
...
+@LargeTest
+@RunWith(AndroidJUnit4.class)
+public class LastUrlTest {
+
+	@Rule
+	public ActivityScenarioRule<SettingsDialogActivity> mActivityTestRule =
+			new ActivityScenarioRule<>(SettingsDialogActivity.class);
+
+	@Test
+	public void somethingToTest() throws InterruptedException {
```
garethbowen, line 45: Let's improve the naming here.

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/LastUrlTest.java
```diff
@@ -0,0 +1,79 @@
...
+				.perform(webClick());
+		onWebView()
+				.withNoTimeout()
+				.withElement(findElement(Locator.ID, "books"))
+				.check(webMatches(getText(), containsString("Books")));
+		Thread.sleep(2000);
+	}
+
+	@Test
+	public void somethingToTest2() throws InterruptedException {
```
garethbowen, line 65: Ditto!

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/LastUrlTest.java
```diff
@@ -0,0 +1,79 @@
...
+		DataInteraction linearLayout = onData(anything())
+				.inAdapterView(allOf(withId(R.id.lstServers),
+						TestUtils.childAtPosition(
+								withId(android.R.id.content),
+								0)))
+				.atPosition(3);
```
garethbowen, line 71: This needs a comment or a better variable name - I have no idea what it is!

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/LastUrlTest.java
```diff
@@ -0,0 +1,79 @@
...
+		ViewInteraction textAppUrl = onView(withId(R.id.txtAppUrl));
+		textAppUrl.perform(replaceText("https://angular.io/resources?category=development"),
+				closeSoftKeyboard());
+		onView(withId(R.id.btnSaveSettings)).perform(click());
+		Thread.sleep(2000);
+		// Click into "Education" to go to "?category=education"
+		onWebView()
+				.withNoTimeout()
+				.withElement(findElement(Locator.XPATH, "//a[contains(text(), 'Education')]"))
+				.perform(webClick());
```
garethbowen, line 56: We probably don't need to click on a new page, right? If the app remembers the original angular.io URL then the code is working.

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -163,6 +168,16 @@ protected void onStart() {
 		super.onStart();
 	}
 
+	@Override
+	protected void onStop() {
+		super.onStop();
+		try {
+			settings.setLastUrl(container.getUrl());
```
garethbowen, line 175: Do you think we need to do any protection to make sure it's still going to the right domain? For example, if someone clicks a link that takes them to google.com, then closes and opens the app, there will be no obvious way to go back to the CHT. It might be worth checking that the domain of the start URL matches the domain of the last URL...

garethbowen (changes requested):
Comments inline.

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/LastUrlTest.java
mrsarm: My bad , I forgot to change that  :facepalm:

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/LastUrlTest.java
mrsarm: It's the position in the grid when tapping which URL to load in the app, going to move the number to a constant and make it more clear...

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
mrsarm: Good point ! Do you think that maybe we should also record the timestamp when was recorded the last URL? so if the recording was time ago we just ignore it, because it may be confusing to enter the app a week later and the app going straight to the last form visited, right? maybe only reload the last URL if the form was recorded the last 24 hours or less.

src/androidTestUnbrandedWebviewDebug/java/org/medicmobile/webapp/mobile/LastUrlTest.java
mrsarm: The first interaction with the UI is to set the CHT URL that is stored only once in the app storage (Unbranded flavor), and after that we need to go to a different URL to verify that when the app is reopened, instead of loading the CHT URL, it loads the last URL opened.

garethbowen: Ahh right. Makes sense. Thanks for the explanation!

src/webview/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
garethbowen: Maybe... I'm not too worried about it because as long as you're still in the CHT then it's easy to get to the page you want. But if you think that'd be an improvement, go for it!

mrsarm:
@garethbowen I addressed your comments, and added a 24 hours limit to load the last URL, please take a look again.

src/main/java/org/medicmobile/webapp/mobile/Utils.java
```diff
@@ -31,6 +33,27 @@ static boolean isUrlRelated(String appUrl, String uriToTest) {
...
+		String parsedBaseUrl = parsedAppUrl.getProtocol() + "://" + parsedAppUrl.getHost();
+		if (navUrl != null
+				&& navUrl.startsWith(parsedBaseUrl)
+				&& !navUrl.contains("/login")
+				&& !navUrl.contains("/_rewrite")) {
+			return true;
+		}
+		return false;
+	}
+
```
garethbowen, line 56: The `isUrlRelated` method above is, as far as I can tell, trying to do a similar thing as this function. They're not completely the same because this checks for `null` and `login` and `_rewrite`, but the regex/parse bit is the same. Find a way to reuse the code so they behave consistently.

mrsarm: I made the refactor to reuse the code, and moved the null checks to that method to make it safer. Could you please take a look again?

garethbowen (approved):
Nice one! Thanks.

Issue #211, opened 2021-09-15
Release v0.10.0


mrsarm:
Release new version [v0.10.0](https://github.com/medic/cht-android/milestone/3).

This version only includes the new [App Launcher integration](https://github.com/medic/cht-android/pull/199).

### Tasks

- [x] Release an alpha version → https://github.com/medic/cht-android/releases/tag/untagged-ff363627098c36503155 (`v0.10.0-alpha.1`)
- [x] Release under the testing channels in the Play Store → Link to join in the open testing channel: https://play.google.com/apps/testing/org.medicmobile.webapp.mobile
- [x] AT with the alpha version and an environment with the changes and configurations needed in the CHT → The instance https://testing.dev.medicmobile.org can be used for this purpose
- [x] Release final version on GH 
- [x] Release final version on the Play Store
- [x] Approved in the Play Store by Google
- [x] Announce the release in the forum.

CC @latin-panda @mrjones-plip @jkuester

newtewt:
I had uninstalled both apps, reinstalled both apps. Launch cht and logged in. The task workflow starts at 59 seconds. At 1:06 you see the blank CHT Android background and then it reload the webapp. 

Please excuse the jumpyness of the whole video. As I had to stitch a couple clips together. Also, you may notice the time is off but this also happens if the time is right and I navigated to a prior tasks that no longer is available in the RDT app because I uninstalled it but you can recreate without clicking into the wrong task. It was just my mistake. 


https://user-images.githubusercontent.com/10531066/133833166-358362e4-7c72-4f5b-947c-1dc3aab2837d.mp4

latin-panda:
Thank you @newtewt @mrjones-plip and @mrsarm for researching this issue, I think a take from the thread discussion is that Android is closing CHT app to free up resources (memory), highly probable that CHT app enters to a lower priority state as it’s waiting in the background for RDToolkit to return a response.

Is there any other thing that needs attention/work? Is this good to proceed? :) 

Thanks!

newtewt:
Don't think we should block for this. It only occurs on the first go around of installing both apps and it recovers pretty well. We also have an issue logged to improve the UX for this experience.

PR #213, opened 2021-09-20
Release notes v0.10.0

mrsarm:
Add upcoming v0.10.0 to CHANGELOG.

Issue: https://github.com/medic/cht-android/issues/211


latin-panda (approved):
Nice!! Thank you putting together the release so fast, it helps us reaching the target 💪

PR #218, opened 2021-10-01
Update README.md

dianabarsan:


mrsarm:
Hey @dianabarsan , I didn't notice this limitation, but may be because something in the settings of the projects?

I have an small OSS on Github, and I added the same rules to skip CI when only mardown files are affected, and it allows to merge the PR without CI being executed. Here is the related settings configuration in that project:

Github_Settings_Protect_Branch-fs8

Am I missing something?

dianabarsan:
@mrsarm yea, in that UI, there should be like a field where it allows you to select which status checks should be required for merging. You can check what happens if you select any build job that doesn't run when you edit .md files.

dianabarsan:
Maybe this actually works and it just needs a review?

mrsarm (approved):

dianabarsan:
Can you merge it after approval @mrsarm  ?

dianabarsan:
I see 
image

mrsarm:
> Maybe this actually works and it just needs a review?

Approved, but still blocked because jobs didn't run :thinking:

CHT-Core_PR_Build_Expected

dianabarsan:
So ... the "skip" rule doesn't work when there are required status checks that are skipped.

PR #220, opened 2021-10-06
Use xxd instead of basez for base16 encoding

derickl:
[xxd](https://linux.die.net/man/1/xxd) has been available for over 2 decades and is available if most *NIX setups.

Supersedes #215

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ abi = x86
 KEYTOOL = keytool
 OPENSSL = openssl
 JAVA = java
-BASE16 = base16
+XXD = xxd
+XXD_OPTS = -p -c 256
 RM_KEY_OPTS = -i
 APKSIGNER = apksigner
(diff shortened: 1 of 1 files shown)


mrsarm (approved):
Awesome ! I was going to work in another approach using Gradle scripting, but this seems to work pretty well. Actually in the past I used to use `hexdump` to see content in an hex representation and `xxd` is much simpler, so added to my toolbox :smile: 

I have tested in my Ubuntu Linux desktop and it worked, did you tested with OSX to be sure not just the command but also the arguments used are compatible? (to avoid issues like with the `head` command) if yes, just merge it, otherwise please move the ticket to AT so it can be tested by the QA team with a Mac machine before merge it.

derickl:
I ran it on Mac and seems you tested on Linux, we have a 'complete' test

PR #232, opened 2021-11-24
Upgrade Gradle, plugins and test dependencies

mrsarm:
In line with https://github.com/medic/cht-gateway/pull/183 , upgrade Gradle and dependencies so both have most of the same dependencies.

- Gradle ~7.1.1~ → 7.3
- Minor upgrade Gradle plug-in to build Android
- Test dependencies upgrade


latin-panda (approved):
LGTM!

Issue #178, opened 2021-05-24
Spaces in URL on unbranded app cause errors


mrjones-plip:
## Overview

When entering in a URL on the unbranded app and accidentally entering space before or after the URL, errors are caused.

We should likely trim leading and trailing spaces for a better user experience since spaces at those positions in the URL are not legal chars (but are mid URL!)

Tested on 
* Android 11 
* Pixel 3a
* Unbranded 0.8.0 webview

## Error 1
The first error is that we give different messages:
* ` https://example.com` - leading space and NO trailing slash, error is: "must be a valid URL"
* ` https://example.com/` - leading space and YES trailing slash, error is: "URL is not a medic-webapp instance"
* `https://example.com ` - trailing space and NO trailing slash, error is: "unable to contact server"
* `https://example.com/ ` - trailing space and YES trailing slash, error is: "URL is not a medic-webapp instance"

steps to reproduce:
1. download unbranded android app
2. enter url with leading or trailing space
3. click "save"

expected: launch app connecting to URL
actual: errors per above

## Error 2
After entering a URL with a space at the beginning and no trailing slash, the launch screen now has a URL entry that when you click the app crashes

steps to reproduce:
1. download unbranded android app
2. enter url with leading space and no trailing slash: ` https://example.com` 
3. click "save"
4. see error c
5. use back button to to back to launch screen
6. click top entry for the URL you just entered on step two

expected: launch app connecting to URL (ideally), or at least show error URL is not a medic-webapp instance"
actual: app crashes b/c `Attempt to invoke virtual method 'void android.widget.TextView.setError(java.lang.CharSequence)' on a null object reference at org.medicmobile.webapp.mobile.SettingsDIalogActivity.a(SettingsDialogActivity.java:8)` (sorry - manually transcribing from screenshot - there may be typos here - see attached)

## Screenshots

![error1](https://user-images.githubusercontent.com/8253488/119406722-08958680-bc98-11eb-9fe7-780bf2c01a78.png)
![error2](https://user-images.githubusercontent.com/8253488/119406724-092e1d00-bc98-11eb-9f6e-baac25d6db23.png)
![error3](https://user-images.githubusercontent.com/8253488/119406726-092e1d00-bc98-11eb-9772-e2b8e2b6052f.png)
![error6](https://user-images.githubusercontent.com/8253488/119406728-09c6b380-bc98-11eb-8908-939ed838952c.png)
[...]

PR #233, opened 2021-11-25
178 - Allow users to input CHT URLs with leading or trailing spaces and or "/"

mrsarm:
User may enter the URL with blank spaces at the beginning or the end while copy & pasting from a message or from the browser. Same with the "/" char at the end, if the user copy & paste the URL from the the URL bar, it may include the "/" at the end that doesn't change the actual resource it wanted to reach.

Issue: #178


src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -29,19 +29,22 @@
...
 	private static final int STATE_LIST = 1;
 	private static final int STATE_FORM = 2;
-
 	private SettingsStore settings;
 	private ServerRepo serverRepo;
 	private int state;
 
 	private static class AppUrlVerificationTask extends ActivityBackgroundTask<SettingsDialogActivity, String, Void, AppUrlVerification> {
+
+		private AppUrlVerifier verifier = new AppUrlVerifier();
```
jkuester, line 38:
```suggestion
		private final AppUrlVerifier verifier = new AppUrlVerifier();
```

src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -29,19 +29,22 @@
...
+
 		AppUrlVerificationTask(SettingsDialogActivity a) {
 			super(a);
 		}
 
 		protected AppUrlVerification doInBackground(String... appUrl) {
 			if(DEBUG && appUrl.length != 1) throw new IllegalArgumentException();
-			return new AppUrlVerifier().verify(appUrl[0]);
+			String chtUrl = verifier.clean(appUrl[0]);
```
jkuester, line 46: It feels like the `clean` functionality should just be built into the `verify` call (so that `verify` just calls `clean` internally and consumers of `AppUrlVerifier` do not have to call both methods).  Then `AppUrlVerifier.clean` can be a private method.

Seems like the only thing this really affects is that for the tests, you will have to assert like this:

```.js
assertEquals("https://example.com/uri", verifier.verify("https://example.com/uri").appUrl);
```
but I don't think that should be a problem.

jkuester (changes requested):
Looks good. Just a few minor comments.   +1 for a full range of unit tests!

src/test/java/org/medicmobile/webapp/mobile/AppUrlVerifierTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.medicmobile.webapp.mobile.R.string.errAppUrl_appNotFound;
+import static org.medicmobile.webapp.mobile.R.string.errAppUrl_serverNotFound;
+import static org.medicmobile.webapp.mobile.R.string.errInvalidUrl;
+import static org.junit.Assert.*;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(sdk=28)
+public class AppUrlVerifierTest {
+
+	private AppUrlVerifier buildAppUrlVerifier(JSONObject jsonResponse) {
```
jkuester, line 21: When possible I like to avoid having to hack like this with `protected` methods.  Java does not make this easy, but I think in this case we could use Mockito to inject a mock SimpleJsonClient2 into the AppUrlVerifier via a second constructor instead of overwriting the protected methods.  The first thing we would have to do is make the json client a class-level variable in AppUrlVerifier (and get rid of `getJsonClient()`) that is set by the constructors:

```.java
	private final SimpleJsonClient2 jsonClient;

	AppUrlVerifier(SimpleJsonClient2 jsonClient) {
		this.jsonClient = jsonClient;
	}

	public AppUrlVerifier() {
		this(new SimpleJsonClient2());
	}
```

Then, this test code can be updated to look something like this:

```.java
	private AppUrlVerifier buildAppUrlVerifier(JSONObject jsonResponse) {
		try {
			SimpleJsonClient2 mockJsonClient = Mockito.mock(SimpleJsonClient2.class);
			Mockito.when(mockJsonClient.get((String) ArgumentMatchers.any())).thenReturn(jsonResponse);
			return new AppUrlVerifier(mockJsonClient);
		} catch (Exception e) {
			throw new RuntimeException(e);
		}
	}

	private AppUrlVerifier buildAppUrlVerifierWithException(Exception e) {
		try {
			SimpleJsonClient2 mockJsonClient = Mockito.mock(SimpleJsonClient2.class);
			Mockito.when(mockJsonClient.get((String) ArgumentMatchers.any())).thenThrow(e);
			return new AppUrlVerifier(mockJsonClient);
		}  catch (Exception exception) {
			throw new RuntimeException(exception);
		}
	}

	private AppUrlVerifier buildAppUrlVerifierOk() {
		try {
			return buildAppUrlVerifier(new JSONObject("{\"ready\":true,\"handler\":\"medic-api\"}"));
		} catch (JSONException e) {
			throw new RuntimeException(e);
		}
	}
```

jkuester (changes requested):

src/main/java/org/medicmobile/webapp/mobile/AppUrlVerifier.java
```diff
@@ -41,6 +45,22 @@ public AppUrlVerification verify(String appUrl) {
...
 	}
+
+	/**
+	 * Clean-up the URL passed, removing leading and trailing spaces, and trailing "/" char
+	 * that the user may input by mistake.
+	 */
+	protected String clean(String appUrl) {
+		appUrl = appUrl.trim();
+		if (appUrl.endsWith("/")) {
+			return appUrl.substring(0, appUrl.length()-1);
```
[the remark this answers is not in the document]
mrsarm: Yes, a trailing "/" is likely, specially when the user copy and paste the URL from an address bar, but I don't see a case where 2 "/" can be introduced, except mistyping, but most of the user just copy and paste the URL from an email, message, web page...

src/test/java/org/medicmobile/webapp/mobile/AppUrlVerifierTest.java
```diff
@@ -0,0 +1,161 @@
...
+	@Test
+	public void testAllMistakesUrl() {
+		AppUrlVerifier verifier = buildAppUrlVerifierOk();
+		AppUrlVerification verification = verifier.verify(" https://example.com/uri/res/ ");
+		assertTrue(verification.isOk);
+		assertEquals("https://example.com/uri/res", verification.appUrl);
+	}
+
+	@Test
+	public void testUrlMalformed() {
```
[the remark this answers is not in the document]
mrsarm: Wrong copy & paste, removing...

src/test/java/org/medicmobile/webapp/mobile/AppUrlVerifierTest.java
mrsarm: I like the idea of injecting the Json client (Spring style :wink: ). I don't like Mockito too much, I think it overrides behaviors that the language already offer (like overriding behavior and injecting objects), allowing developers to write code with anti-patterns that the other way is less possible, because that kind of code is less testable unless... mockito.

Anyway, we already have Mockito in the project and it makes the code more compact in many cases (much appreciated when you write in Java), so I'll use it... against my will :stuck_out_tongue:

jkuester: Honestly, Mockito always makes me cringe too. Every time I use it I just get to thinking "there has to be a better way".  But, I have never been able to find one (unless your production code is already structured to leverage some kind of dependency injection or service provider framework).  

In this case it feels like we are using Mockito the "good" way in that we are just using it to construct objects with mocked behavior that we can inject (and not doing any kind of deep manipulation on the implementation code...).

jkuester (approved):

PR #240, opened 2021-12-15
636 test muso cdi

adrienlemec:
[#636](https://github.com/medic/config-muso/issues/636)


binokaryg:
Looks like you still have a conflict. Can you please resolve it?

adrienlemec:
@binokaryg conflicts resolved!

binokaryg:
You don't have icons added for the app logo?

src/muso_cdi_test/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string name="app_name">Muso CDI Test</string>
+  <string name="app_host">muso-cdi.dev.medicmobile.org</string>
```
Marina-Kenf, line 4: Great job here @adrienlemec.
Are we planning to create production's ( `muso-cdi.app.medicmobile.org` ) flavor at another stage?

binokaryg: Oh, I missed that it was pointing to the dev URL. I don't think it's necessary to create an official APK with this method pointing to the dev instance. We can easily create it locally and share it via email, drive etc. Also, the number of flavors is already too high, let's not make it even more crowded.

Thanks for catching it, Marina.

adrienlemec: Alright, I have the APK locally. Let me share it with everyone.

adrienlemec:
> You don't have icons added for the app logo?

We're waiting Hiell/Fatou to provide the App name and the Logo

Marina-Kenf:
@ernestoteo looks like Hiell has shared details on the app's name and Logo. Could you plan to integrate it in this PR? cc @adrienlemec

ernestoteo:
Thanks @Marina-Kenf , the changes are made.

src/muso_cdi_test/res/values/strings.xml
binokaryg: Do we still want to merge it or the APK created locally would suffice?

binokaryg:
> Do we still want to merge it or the APK created locally would suffice?

@adrienlemec ?

adrienlemec:
> > Do we still want to merge it or the APK created locally would suffice?
> 
> @adrienlemec ?

I don't know. But, I think we should follow the normal steps of APK creation (by merging, in fact). 

@Marina-Kenf what do you think?

binokaryg:
> I think we should follow the normal steps of APK creation (by merging, in fact)

If we look at the [build.gradle](https://github.com/medic/cht-android/blob/636-test-muso-cdi/build.gradle) or [publish.yml](https://github.com/medic/cht-android/blob/636-test-muso-cdi/.github/workflows/publish.yml) file, we don't see many `test` or `dev` flavors. Is it worth the time it takes to assemble and run the actions for a test app on each change?

Marina-Kenf:
I think this PR will no more be useful as Muso's developer finalized [this one](https://github.com/medic/cht-android/pull/248), already merged to master. I believe we can safely close this one.

Issue #230, opened 2021-11-18
Remove simprints code


garethbowen:
Nobody is using it and it's broken right now, so we should remove it to reduce bloat. We can always add it back when someone needs it.

latin-panda:
Ready for AT in the following PRs:

PR CHT-Core:
medic/cht-core#7461
CHT-Android:
medic/cht-android#241
PR CHT-Docs:
medic/cht-docs#598

Android app and cht-core should work correctly after removing Simprints.

ngaruko:
CHT and cht-android work. There seems to be a few references to simprints in some [translations files](https://github.com/medic/cht-core/blob/230-bye-simprints/ddocs/medic/_attachments/translations/messages-es.properties#L1180), but since they are not shipped with code, we might need a different ticket in the future for this, maybe to update entries in POEditor.
Please feel free to merge @latin-panda.

latin-panda:
Thanks @ngaruko, the translation's updates are already loaded in POEditor.

<img width="300" alt="Screen Shot 2022-01-14 at 8 14 44 am" src="https://user-images.githubusercontent.com/66472237/149529341-d2f68f0e-efba-4da3-a05f-6353d72982e5.png">

I'll merge this one once the x-walk clean up is merged as well.

PR #241, opened 2022-01-07
230 - Removing simprints

latin-panda:
# Description

This PR removes Simprints integration.

Ticket:
https://github.com/medic/cht-android/issues/230
PR CHT-Core:
https://github.com/medic/cht-core/pull/7461
PR CHT-Docs:
https://github.com/medic/cht-docs/pull/598

# License


latin-panda:
Hi @jkuester, can you please have a look? Thanks!

build.gradle
```diff
@@ -30,8 +30,6 @@ repositories {
   }
```
jkuester, line 30: Can we remove the `jcenter()` repository now?  I think that is where we were loading the com.simprints:LibSimprints dependency.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -488,4 +442,33 @@ private void registerRetryConnectionBroadcastReceiver() {
 		};
 		registerReceiver(broadcastReceiver, new IntentFilter("retryConnection"));
 	}
+
+//> CLASSES
+	public static class RequestCode {
```
jkuester, line 447: Minor, but this looks like a pretty good place to use a Java enum:

```.java
	public enum RequestCode {
		ACCESS_LOCATION_PERMISSION(100),
		ACCESS_STORAGE_PERMISSION(101),
		CHT_EXTERNAL_APP_ACTIVITY(102),
		DISCLOSURE_LOCATION_ACTIVITY(103),
		GRAB_MRDT_PHOTO_ACTIVITY(104),
		GRAB_PHOTO_ACTIVITY(105);

		private final int requestCd;

		RequestCode(int requestCd) {
			this.requestCd = requestCd;
		}

		public static Optional<RequestCode> valueOf(int requestCd) {
			return Arrays.stream(RequestCode.values()).filter(e -> e.getCode() == requestCd).findFirst();
		}

		public int getCode() {
			return requestCd;
		}
	}
```

Then above you can just switch directly on the enum itself.  (Also, if you need the string value to trace out, you can just use the `RequestCode.name()` method.)

jkuester (changes requested):
Very nice!  Couple minor suggestions inline.

build.gradle
latin-panda: Not yet, we need it for "com.noveogroup.android:check" dependency that makes static check. 
Have you used another library for the same checks?

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: Nice suggestion! It's done

latin-panda:
@jkuester thanks for the feedback! I've uploaded the requested changes :)

build.gradle
jkuester: LOL, went looking for why the latest version of android:check was not in mavenCentral and found [this issue](https://github.com/noveogroup/android-check/issues/43) logged by someone familiar....

Seems like we need to keep the jcenter() repo for now, but since android:check is unmaintained now, I have logged a new issue to migrate off of that plugin: https://github.com/medic/cht-android/issues/242

latin-panda: Thank you! :)

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -192,38 +193,39 @@ protected void onStop() {
 				backButtonHandler);
 	}
 
-	@Override protected void onActivityResult(int requestCode, int resultCode, Intent intent) {
-		String requestCodeLabel = RequestCode.requestCodeToString(requestCode);
+	@Override protected void onActivityResult(int requestCd, int resultCode, Intent intent) {
+		RequestCode requestCode = RequestCode.valueOf(requestCd).get();
```
jkuester, line 197: We should probably check the optional before just calling `get()` (since the code here before at least seems to consider the possibility of an invalid requestCd).  Could do something like: 

```.java
	@Override protected void onActivityResult(int requestCd, int resultCode, Intent intent) {
		Optional<RequestCode> requestCodeOpt = RequestCode.valueOf(requestCd);
		if(requestCodeOpt.isPresent()) {
			RequestCode requestCode = requestCodeOpt.get();
			try {
				trace(this, "onActivityResult() :: requestCode=%s, resultCode=%s", requestCode.name(), resultCode);

				switch (requestCode) {
					case GRAB_PHOTO_ACTIVITY:
						photoGrabber.process(requestCode, resultCode, intent);
						return;
					case GRAB_MRDT_PHOTO_ACTIVITY:
						processMrdtResult(requestCode, resultCode, intent);
						return;
					case DISCLOSURE_LOCATION_ACTIVITY:
						processLocationPermissionResult(resultCode);
						return;
					case CHT_EXTERNAL_APP_ACTIVITY:
						processChtExternalAppResult(resultCode, intent);
						return;
					default:
						trace(this, "onActivityResult() :: no handling for requestCode=%s", requestCode.name());
				}
			} catch(Exception ex) {
				String action = intent == null ? null : intent.getAction();
				warn(ex, "Problem handling intent %s (%s) with requestCode=%s & resultCode=%s",
					intent, action, requestCode.name(), resultCode);
			}
		} else {
			trace(this, "onActivityResult() :: no handling for requestCode=%s", requestCd);			
		}
	}
```

(Sadly this is not quite as clean as it should be since it seems like our version of the Android SDK does not support the Optional.ifPresentOrElse(...) method...)

jkuester (changes requested):

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: Thanks! Done

jkuester (approved):

Issue #242, opened 2022-01-11
Migrate off of the unmaintained com.noveogroup.android:check plugin


jkuester:
**Describe the issue**
Currently we are using the `com.noveogroup.android:check` plugin to do static analysis as part of the Gradle build.  Unfortunately, this plugin is not being maintained anymore ([no commits](https://github.com/noveogroup/android-check) since 2017). Also, the latest version of this plugin (`1.2.5`) appears to be only available from the `jcenter()` repository which [is deprecated](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/).

**Describe the improvement you'd like**
I believe that this plugin really only serves as a wrapper for [checkstyle](https://docs.gradle.org/current/userguide/checkstyle_plugin.html), [PMD](https://docs.gradle.org/current/userguide/pmd_plugin.html), and [SpotBugs](https://spotbugs.readthedocs.io/en/latest/gradle.html).  Perhaps we can just use those plugins directly.  This will probably require more config in our repo, but we will have the benefit of more up-to-date versions of these plugins.

PR #244, opened 2022-01-31
242 - Replacing the old static checks library

latin-panda:
# Description
The library for static checks is unmaintained `com.noveogroup.android:check`.
In this PR:
- Replaced this library with the latest version of the libraries it was wrapping: Checkstyles, PMD and Spotbugs.
- Removed `jcenter()` which will stop working at some point in the future. 

Ticket: https://github.com/medic/cht-android/issues/242

# License


build.gradle
```diff
@@ -1,17 +1,20 @@
 buildscript {
   repositories {
     mavenCentral()
-    maven { url "https://jitpack.io" }
+    maven { url 'https://jitpack.io' }
+    maven { url 'https://plugins.gradle.org/m2/' }
```
latin-panda, line 5: Latest version of spotbugs for gradle is available in `https://plugins.gradle.org/m2/`

build.gradle
```diff
@@ -20,10 +23,35 @@ gradle.projectsEvaluated {
...
+  group 'verification'
+  configFile file('config/checkstyle.xml')
+  source 'src'
+  classpath = files()
+  ignoreFailures = false
+}
+
+task pmd(type: Pmd) {
+  ruleSetFiles = files('config/pmd.xml')
+  ruleSets = []
```
latin-panda, line 37: It's an empty array because otherwise it'll ignore the `config/pmd.xml` that has our config.

config/checkstyle.xml
```diff
@@ -8,7 +8,7 @@
 	<!-- Checks whether files end with a new line. -->
 	<!-- See http://checkstyle.sourceforge.net/config_misc.html#NewlineAtEndOfFile -->
 	<module name="NewlineAtEndOfFile">
-		<property name="severity" value="info"/>
+		<property name="severity" value="warning"/>
```
latin-panda, line 11: Or "ignore", thoughts?

src/covid_moh_mali/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -2,4 +2,4 @@
 <adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
     <background android:drawable="@mipmap/ic_launcher_background"/>
     <foreground android:drawable="@mipmap/ic_launcher_foreground"/>
-</adaptive-icon>
\ No newline at end of file
+</adaptive-icon>
```
latin-panda, line 5: So checkstyle throws a warning/info message referring that these files don't have a new line at the end of the file.

latin-panda:
Hi @jkuester Can you please have a look and let me know your thoughts on these checks? Thanks!

jkuester:
@latin-panda one question I have run into while going through these changes is how are you running the `spotbugs` plugin/task?  When I try running `gradle spotbugsUnbrandedRelease`, for example, I get an error: 

```
> Task :spotbugsUnbrandedRelease
Scanning archives (101 / 101)
2 analysis passes to perform
Pass 1: Analyzing classes (356 / 356) - 100% complete
Pass 2: Analyzing classes (13 / 64) - 20% completeM P UPM: Private method org.medicmobile.webapp.mobile.SimpleJsonClient2.log(Exception, String, Object[]) is never called  At SimpleJsonClient2.java:[lines 163-164]
Pass 2: Analyzing classes (63 / 64) - 98% complete
Done with analysis
The following classes needed for analysis were missing:
  java.rmi.Remote

> Task :spotbugsUnbrandedRelease FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spotbugsUnbrandedRelease'.
> A failure occurred while executing com.github.spotbugs.snom.internal.SpotBugsRunnerForHybrid$SpotBugsExecutor
   > Verification failed: SpotBugs ended with exit code 3
```

latin-panda:
@jkuester you're running the task correctly. Spotbugs is complaining that there's [1 method](https://github.com/medic/cht-android/blob/d4df8eb4e887be56d89eeb6b0dad3703128af115/src/main/java/org/medicmobile/webapp/mobile/SimpleJsonClient2.java#L162) that isn't used 
however this method is actually called for [the debug flavour only.](https://github.com/medic/cht-android/blob/d4df8eb4e887be56d89eeb6b0dad3703128af115/src/main/java/org/medicmobile/webapp/mobile/SimpleJsonClient2.java#L112) 😩 that's the check passes on green for ./gradlew spotbugsUnbrandedDebug but in red for ./gradlew spotbugsUnbrandedRelease
I'll suppress the check for this specific case.

latin-panda:
@jkuester I've sent the commit for this case.

jkuester:
> Spotbugs is complaining that there's 1 method that isn't used

Hah, wow :facepalm: I totally missed that part of the log.  Just got focused on `The following classes needed for analysis were missing:  java.rmi.Remote`...  (which after doing some more reading is apparently is not actually a big deal).  Thanks for the update!

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -291,7 +296,11 @@ public String getDeviceInfo() {
 					.put("ram", ramObject)
 					.put("network", networkObject)
 					.toString();
-		} catch(Exception ex) {
+		} catch (RuntimeException e) {
```
jkuester, line 299: I think we actually might want to just suppress the warning here.  From the docs:

>It is a common bug pattern to say try { ... } catch (Exception e) { something } as a shorthand for catching a number of types of exception each of whose catch blocks is identical, but this construct also accidentally catches RuntimeException as well, masking potential bugs.

But, this is not what this code is actually trying to do.  If I understand this correctly we never want to throw any exceptions (including `RuntimeException`s) out of this method, but instead should always be logging the exception and returning the `jsonError` (so that downstream code can handle the error).

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
```diff
@@ -328,6 +337,8 @@ public void onDateSet(DatePicker view, int year, int month, int day) {
 		dialog.show();
 	}
 
+	@SuppressFBWarnings("DM_DEFAULT_ENCODING")
+	@SuppressWarnings("PMD.CloseResource")
```
jkuester, line 341: Instead of suppressing this warning, it is probably better to wrap everything in a try-with-resources block so that both the readers will always be closed (even if an exception gets thrown):

```.java
		try(
			FileReader fr = new FileReader("/proc/cpuinfo");
			BufferedReader bufferedReader = new BufferedReader(fr);
		){
			String line;
			HashMap output = new HashMap();
			while ((line = bufferedReader.readLine()) != null) {
				String[] data = line.split(":");
				if (data.length > 1) {
					String key = data[0].trim();
					if (key.equals("model name")) {
						output.put(key, data[1].trim());
						break;
					}
				}
			}

			int cores = Runtime.getRuntime().availableProcessors();
			output.put("cores", cores);

			String arch = System.getProperty("os.arch");
			output.put("arch", arch);

			return output;
		}
```

src/main/java/org/medicmobile/webapp/mobile/SimpleJsonClient2.java
```diff
@@ -37,6 +39,7 @@ public JSONObject get(String url) throws JSONException, IOException {
 		return get(new URL(url));
 	}
 
+	@SuppressWarnings("PMD.CloseResource")
```
jkuester, line 42: I think the only thing wrong here is that PMD is not smart enough to realize that the `closeSafely` method is actually closing these resources.  The [`CloseResource`](https://pmd.github.io/latest/pmd_rules_java_errorprone.html#closeresource) PMD rule actually allows for configuring this to let it know that custom methods handle the closing.  I think we just need to add this to our `pmd.xml`: 

```.xml
  <rule ref="category/java/errorprone.xml/CloseResource">
    <properties>
      <property name="closeTargets" value="closeSafely" />
    </properties>
  </rule>
```

src/main/java/org/medicmobile/webapp/mobile/SimpleJsonClient2.java
```diff
@@ -158,6 +161,7 @@ private static void log(String methodName, String message) {
 		Log.d(LOG_TAG, "SimpleJsonClient2." + methodName + "() :: " + message);
 	}
 
+	@SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD")
```
jkuester, line 164:
```suggestion
	@SuppressFBWarnings(value = "UPM_UNCALLED_PRIVATE_METHOD", justification = "Only called in Debug flavor")
```

src/main/java/org/medicmobile/webapp/mobile/XWalkMigration.java
```diff
@@ -103,12 +105,14 @@ public void run() {
 		}
 	}
 
+	@SuppressFBWarnings("RV_RETURN_VALUE_IGNORED")
```
jkuester, line 108: I am a little concerned about just suppressing this warning (both here and in the methods below).  The doc for the `File.renameTo` method says:

> The return value should always be checked to make sure that the rename operation was successful.
> ...
> Returns: true if and only if the renaming succeeded; false otherwise

This sounds to me like the renaming can actually fail without causing an exception.  It seems worth considering here if we should actually be checking the return value and throwing an error here if we could not rename the file.  What do you think?

I guess it comes down to the question of whether we should continue starting the app if the xWalk migration fails. (But, I am not confident that I know enough about the workflow to make a solid recommendation here...)

jkuester (dismissed):
Good work!  It is nice to see this getting done so fast!  Couple of questions and suggestions inline....

build.gradle
```diff
@@ -20,10 +23,35 @@ gradle.projectsEvaluated {
   }
 }
 
+task checkstyle(type: Checkstyle) {
+  description 'Check code standard'
+  group 'verification'
+  configFile file('config/checkstyle.xml')
+  source 'src'
+  classpath = files()
```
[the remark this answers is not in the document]
latin-panda: Sure, the `source` is [here](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.Checkstyle.html#org.gradle.api.plugins.quality.Checkstyle:source) and the one for pmd is [here](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.Pmd.html#org.gradle.api.plugins.quality.Pmd:source). Maybe something standard or inherit from a base class.

config/checkstyle.xml
latin-panda: Agree. I'm going to add `ignore` to `NewlineAtEndOfFile` and `error` for trailing spaces.

config/pmd.xml
```diff
@@ -1,117 +1,25 @@
...
-Severity: HARD
--->
-<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="ruleset"
-	xmlns="http://pmd.sf.net/ruleset/1.0.0"
-	xsi:noNamespaceSchemaLocation="http://pmd.sf.net/ruleset_xml_schema.xsd"
-	xsi:schemaLocation="http://pmd.sf.net/ruleset/1.0.0 http://pmd.sf.net/ruleset_xml_schema.xsd">
-
-	<description>POM rule set file</description>
-
-	<rule ref="rulesets/java/android.xml"/>
```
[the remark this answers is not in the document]
latin-panda: Well the `check` library takes this config and use it instead, _I think_ it doesn't add on top. 
Some of these checks doesn't exists anymore and some got moved to another place (mainly to errorprone)
I was able to map some of these in the links provided in [this comment](https://github.com/pmd/pmd/issues/805#issuecomment-353768976),  like for example [AssignmentInOperand](https://github.com/pmd/pmd/blob/81739da5caff948dbcd2136c17532b65c726c781/pmd-java/src/main/resources/rulesets/java/controversial.xml#L14) was moved from `rulesets/java/controversial` to `errorprone`.

The pmd task default [set of rules](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.PmdExtension.html#org.gradle.api.plugins.quality.PmdExtension:ruleSets) is `errorprone`, like the minimum suggested. I included this one in the config xml and turned off some rules that we're turning off before in the original config. 
I noticed that [Spotbugs](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#upm-private-method-is-never-called-upm-uncalled-private-method) catches some of the pmd's [additional rules](https://pmd.github.io/latest/pmd_rules_java_bestpractices.html). 

Is there any in particular that we should include in the pmd?

src/main/java/org/medicmobile/webapp/mobile/LockScreen.java
```diff
@@ -61,6 +63,7 @@
 		final ViewGroup group = (ViewGroup) findViewById(R.id.divButtons);
 		int i = group.getChildCount();
 		OnClickListener buttonListener = new OnClickListener() {
+			@SuppressFBWarnings("BC_UNCONFIRMED_CAST")
```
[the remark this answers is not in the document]
latin-panda: oh woooow magic 🤩 love it
I didn't thought about adding `assert` in the actual code (outside the test cases).
Done!

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
latin-panda: 👍Done

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
latin-panda: Nice done!

latin-panda: I also got another error in spotbugs: `Found reliance on default encoding`
So end up making the code like from this:
```
FileReader fr = new FileReader("/proc/cpuinfo");
```
to:
```
Reader fileReader = new InputStreamReader(new FileInputStream("/proc/cpuinfo"), StandardCharsets.UTF_8);
			BufferedReader bufferedReader = new BufferedReader(fileReader);
```

src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -260,26 +260,30 @@ void save(String url) {
...
 		if(url.endsWith(".medicmobile.org/")) {
 			url = url.substring(0, url.length() - ".medicmobile.org/".length());
 		}
+
 		if(url.startsWith("192.168.")) {
 			return url.substring("192.168.".length());
 		} else {
 			String[] parts = url.split("\\.");
-			url = "";
+			StringBuffer stringBuffer = new StringBuffer();
```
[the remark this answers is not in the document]
latin-panda: Thanks!

src/main/java/org/medicmobile/webapp/mobile/SimpleJsonClient2.java
latin-panda: Thanks! done

src/main/java/org/medicmobile/webapp/mobile/SimpleJsonClient2.java
latin-panda: 👍

src/main/java/org/medicmobile/webapp/mobile/XWalkMigration.java
latin-panda: Same I'm not sure 🤔 maybe the original intention is to ignore the outcome because not significant and just some phones needs to do the migration. I'm going to log a message when it couldn't rename or delete the files just in case.

latin-panda: I don't think it should stop the start of the app, since it moves the web resources like local storage, so even if it fails to move these resources from XWalk to WebView, the user should be able to use the app in WebView, just that they might need to download the resource and cache again.

latin-panda: Warnings added.

latin-panda:
Thanks for the great feedback @jkuester, I've applied it. 

Hi @garethbowen, can you please review of the static check config since Josh will be out? Thank you for your time :)

build.gradle
```diff
@@ -1,17 +1,20 @@
...
+    classpath 'com.android.tools.build:gradle:7.1.0'
+    classpath 'com.github.spotbugs.snom:spotbugs-gradle-plugin:5.0.5'
   }
 }
+
 apply plugin: 'com.android.application'
-apply plugin: 'com.noveogroup.android.check'
+apply plugin: 'checkstyle'
+apply plugin: 'pmd'
+apply plugin: "com.github.spotbugs"
```
garethbowen, line 17:
```suggestion
apply plugin: 'com.github.spotbugs'
```

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
garethbowen: I think you still need to close the resources in a `finally` block. In other words, we shouldn't just ignore this error.

src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -260,26 +260,30 @@ void save(String url) {
...
 			for(String p : parts) {
-				url += " ";
-				url += p.substring(0, 1).toUpperCase();
-				url += p.substring(1);
+				stringBuilder.append(" ");
+				stringBuilder.append(p.substring(0, 1).toUpperCase());
+				stringBuilder.append(p.substring(1));
 			}
-			return url.substring(1);
+			return stringBuilder.toString().substring(1);
```
garethbowen, line 286: Much better, thanks!

src/main/java/org/medicmobile/webapp/mobile/UrlHandler.java
```diff
@@ -17,10 +17,13 @@
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 public class UrlHandler extends WebViewClient {
 	EmbeddedBrowserActivity parentActivity;
 	SettingsStore settings;
 
+	@SuppressFBWarnings("EI_EXPOSE_REP2")
```
garethbowen, line 26: We're suppressing this a lot. Maybe we should just ignore it throughout rather than add all these annotations?

src/main/java/org/medicmobile/webapp/mobile/XWalkMigration.java
```diff
@@ -167,10 +172,21 @@ private File getStorageRootFromFiles(File filesDir) {
...
-			for (File child : fileOrDirectory.listFiles())
+		if (fileOrDirectory == null) {
+			return;
+		}
+
+		if (fileOrDirectory.isDirectory()) {
+			File[] files = fileOrDirectory.listFiles();
+			if (files != null) for (File child : files) {
 				deleteRecursive(child);
+			}
```
garethbowen, line 183: This inline `for` is a little non-standard. How about...

```suggestion
			if (files != null) {
				for (File child : files) {
					deleteRecursive(child);
				}
			}
```

garethbowen (changes requested):
Nice one. I've got a few comments inline.

src/main/java/org/medicmobile/webapp/mobile/MedicAndroidJavascript.java
latin-panda: Oh I forgot to remove the suppress, this is fixed. When opening the resources in the `try()` they'll get closed automatically without needing the `finally`.

build.gradle
latin-panda: Thanks!

src/main/java/org/medicmobile/webapp/mobile/UrlHandler.java
latin-panda: Done, adding a note saying that we omit this rule because we connect with external apps and assume the data exchanged is safe.

src/main/java/org/medicmobile/webapp/mobile/XWalkMigration.java
latin-panda: Done

latin-panda:
Thanks for the feedback @garethbowen, I've pushed the changes. Can you please have a look again? Thanks!

garethbowen (approved):
Awesome! Great work, thanks.

PR #251, opened 2022-03-23
Add compatibility matrix to clarify support

garethbowen:


garethbowen:
@craig-landry Review please!

This is a response to your excellent question about cross compatibility. Thinking it through there's no reason why we can't support cht-core v3.x with cht-android v1.0, and it'll make upgrading much easier if we support upgrading cht-android first and then upgrading cht-core once complete, so I'm proposing we officially support cht-core v3 with cht-android v1. What do you think?

Additionally while this is a decent place to put this info right now, it's probably hard to find. Should this doc be migrated to the docs site (akin to cht-core releases), or somewhere else?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# Compatibility
+
+| CHT Android | Android OS | cht-core |
+| -- | -- | -- |
+| 1.x | 5.0+ | 3.x and 4.x |
```
craig-landry, line 5: Cool, I think this further clarified a question I had yesterday (that I thought I understood)! So if you use Android 1.0, we will test it against cht-core 3.x. Am I understanding that correctly?

craig-landry (approved):
Approving assuming my question/confirmation is correct.

CHANGELOG.md
garethbowen: Yes. When cht-android 1.0 (and future 1.x releases) are being release tested we'll have to test it against cht-core 3.x and 4.x.

Issue #253, opened 2022-03-31
Remove the old menu and lock pin screen


latin-panda:
**Describe the issue**
I found some old code: 

1. We have [a menu](https://github.com/medic/cht-android/blob/e107f89f9873a4fbed52d83e485660c42f1fe6d1/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java#L166) that used to work with a menu button that old phones used to have, [this button was removed 8+ years ago](https://www.androidauthority.com/android-menu-button-is-dead-48604/). 

2. A [lock](https://github.com/medic/cht-android/blob/master/src/main/java/org/medicmobile/webapp/mobile/LockScreen.java) and [pin page](https://github.com/medic/cht-android/blob/master/src/main/res/layout/pin_entry.xml), see the design attached. I’ve never seen / heard this one actively working. I know there were some ideas about possibly adding something like this, but this code is very old, so if we ever want to add a pin then we can write a fresh modern code that is covered with testing.

<img width="300" src="https://user-images.githubusercontent.com/66472237/161080523-5910bff9-c0b2-42a7-bd85-e1b9d5dd9867.png">

**Describe the improvement you'd like**
Remove the menu and lock pin view.
Add a secret door to open the settings view to facilitate testing across instances.

latin-panda:
Ready for AT in [this PR](https://github.com/medic/cht-android/pull/254)

AT recommendation:
- Testing that old devices are working fine even when clicking the old menu button (the one from the device, not the one from webapp) since we have removed the menu.
- Testing that app goes back, closes, open and navigates across tabs correctly. Since we removed that LockActivity because we don't use the lock pin view anymore, and the main activity that opens webapp was depending on this LockActivity. 
- Testing that the ["secret door"](https://github.com/medic/cht-android/pull/254/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R18) opens the Settings view, that going back from Settings view and then go again to Settings view works fine. If you tap on a clickable element, the clickable element have more priority than the "secret door" _and that's expected_, so tap in non-clickable elements, for example in the white background of area of targets or message module.

ngaruko:
Most of the app seems to work. 
However, there is one little glitch: most up will exit when the user presses the back button "too many times", but our app does not seem to exit/_finish()_, even if there are no activities in the back stack.
Is this intentional @latin-panda ?

[...]

PR #254, opened 2022-03-31
253 - Removing the menu and lock pin.

latin-panda:
# Description

This PR: 
- Removes the menu.
- Removes the lock pin view and activity.
- Adds secret door to open the settings view, that used to be available in the menu. Tap 5 times with 1 finger on the screen, then swipe right with 2 fingers. Tap counting will restart if next tap happened more than 500 milliseconds since last tap. 
- Adds test coverage.

Ticket: https://github.com/medic/cht-android/issues/253

# License


src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -66,6 +70,7 @@ public void onReceiveValue(String result) {
 
 
 //> ACTIVITY LIFECYCLE METHODS
+	@SuppressLint("ClickableViewAccessibility")
```
latin-panda, line 73: Because the onTouchEvent() isn't calling View#performClick

latin-panda: Because we don't want to click: https://developer.android.com/reference/android/view/View#onTouchEvent(android.view.MotionEvent)

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -122,6 +128,7 @@ public void onReceiveValue(String result) {
 		}
 	}
 
+	@SuppressWarnings("PMD.CallSuperFirst")
```
latin-panda, line 131: I don't feel confident of playing with the xwalk browser migration in this PR 🤔

latin-panda:
Hi @jkuester, when you have time, can you please review this one too? Thank you!

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: yes, I think is that

src/main/java/org/medicmobile/webapp/mobile/GestureHandler.java
```diff
@@ -0,0 +1,24 @@
...
+
+import android.view.MotionEvent;
+
+public class GestureHandler {
+
+	private float finger1AxisX;
+	private float finger2AxisX;
+	private static final float MIN_SWIPE_LENGTH = 200;
+
+	public GestureHandler(float finger1AxisX, float finger2AxisX) {
```
jkuester, line 11: Minor, but since the `isSwipeRight` method just takes a `MotionEvent` object, is there a reason we cannot just pass the same in here?

src/test/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivityTest.java
```diff
@@ -164,4 +171,168 @@ public void getLocationPermissions_withPermissionsAlreadyDenied_returnsFalse() {
...
+				MotionEvent event = mock(MotionEvent.class);
+				when(event.getPointerCount()).thenReturn(2);
+				when(event.getActionMasked()).thenReturn(MotionEvent.ACTION_POINTER_DOWN);
+				when(event.getX(0)).thenReturn((float) 262.99);
+				when(event.getX(1)).thenReturn((float) 280.49);
+
+				embeddedBrowserActivity.onKeyDown(KeyEvent.KEYCODE_VOLUME_DOWN, null);
+				OnTouchListener onTouchListener = embeddedBrowserActivity.onTouchEvent();
+
+				//>THEN
```
jkuester, line 191: Might just be misunderstanding how you are thinking about things, but it feels like you might have accidentally switched the `WHEN` and `THEN` labels on these tests...

src/test/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivityTest.java
```diff
@@ -164,4 +171,168 @@ public void getLocationPermissions_withPermissionsAlreadyDenied_returnsFalse() {
...
+				//> GIVEN
+				MotionEvent event = mock(MotionEvent.class);
+				when(event.getPointerCount()).thenReturn(2);
+				when(event.getActionMasked()).thenReturn(MotionEvent.ACTION_POINTER_DOWN);
+				when(event.getX(0)).thenReturn((float) 262.99);
+				when(event.getX(1)).thenReturn((float) 280.49);
+
+				embeddedBrowserActivity.onKeyDown(KeyEvent.KEYCODE_VOLUME_DOWN, null);
+				OnTouchListener onTouchListener = embeddedBrowserActivity.onTouchEvent();
+				embeddedBrowserActivity.onKeyUp(KeyEvent.KEYCODE_VOLUME_DOWN, null);
```
jkuester, line 226: Minor, but it might be nice to move this line down to come right before the second call to `onTouch` (to additionally demonstrate that the volume button has to be held down through the entire duration of the swipe gesture).

jkuester (changes requested):
Good work!  I have asked a couple of questions and put a few proposals inline.

One thing that I have been considering is that it feels like we really should document this behavior somewhere.  (I get that it is supposed to be "secret", but for the sake of discoverability, it feels like we should write it down somewhere...). The main issue is that I have not been able to come up with a good place to suggest that we add the docs.  At first I was thinking https://docs.communityhealthtoolkit.org/core/guides/android/, but there does not really seem to be a place out there that fits...

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -330,6 +320,34 @@ public void locationRequestResolved() {
 		evaluateJavascript("window.CHTCore.AndroidApi.v1.locationPermissionRequestResolved();");
 	}
 
+//> PROTECTED
+	OnTouchListener onTouchEvent() {
```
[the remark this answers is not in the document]
latin-panda: I think the idea makes sense. I'm just trying to fix an issue when doing the trick to display Settings Dialog, then press the phone's back button back to EmbeddedBrowserActivity. Then do the trick again but this time the Settings won't show, for some reason the listeners stop working. 
I have to navigate through tabs Messages to Reports, then the trick to open the Settings will work.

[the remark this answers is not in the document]
latin-panda: Looks like the culprit is the OnKeyListener 🤔

src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java
```diff
@@ -41,28 +42,16 @@ public boolean isRootUrl(String url) {
 
 	public abstract void update(SharedPreferences.Editor ed, WebappSettings s);
 
-	String getUnlockCode() {
-		return get("unlock-code");
```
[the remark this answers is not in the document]
latin-panda: It'd be nice to have it clean but 🤔 it represents a very tiny amount of occupied space that maybe not worth having these lines of code for years in our project and running everyone and then. For now I'll skip this cleaning.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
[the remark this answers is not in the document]
latin-panda: I changed the "secret door" to move forward since the OnKeyListener was giving me a lot of trouble with the back button and view scope. 
Now the user will have to tap 5 times with 1 finger anywhere on the screen, then swipe right with 2 fingers. Tap counting will restart if next tap happened more than 500 miliseconds since last tap.

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
```diff
@@ -0,0 +1,82 @@
...
+			return;
+		}
+
+		long currentTime = System.currentTimeMillis();
+		fingerTabCount = lastTimeTab + TIME_BETWEEN_TABS >= currentTime ? fingerTabCount + 1 : 1;
+		lastTimeTab = currentTime;
+	}
+
+	private boolean hasTabEnough() {
+		return fingerTabCount == 6;
```
latin-panda, line 54: 5 tabs by the user + 1 for the swipe.

latin-panda:
Hi @jkuester, can you please review again? Thanks!
At the stand up, I brought up the topic of where to write down the secret door sequence, we thought on adding it in the readme file for now.

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
```diff
@@ -0,0 +1,193 @@
...
+		openSettingsDialogFragment = mock(OpenSettingsDialogFragment.class, fragmentSettings);
+		when(openSettingsDialogFragment.getActivity()).thenReturn(activity);
+		argsStartActivity = ArgumentCaptor.forClass(Intent.class);
+		doNothing().when(openSettingsDialogFragment).startActivity(argsStartActivity.capture());
+
+		openSettingsDialogFragment.onCreate(null);
+	}
+
+	private void tap(OnTouchListener onTouchListener, MotionEvent eventTab, int times) {
+		for (int i = 0; i < times; i++) {
```
jkuester, line 59: I recently learned about the `IntStream.range` method which is essentially an easy way to created a sized Stream.

That would allow for something like: `IntStream.range(0, times).forEach(i -> onTouchListener.onTouch(null, eventTab));` which is handy if you like functional code and want to avoid loops....  (Though, I don't think it really helps to improve the "readability" of the code...)

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
```diff
@@ -0,0 +1,193 @@
...
+		MotionEvent eventSwipe = mock(MotionEvent.class);
+		when(eventSwipe.getPointerCount()).thenReturn(2);
+
+		//> WHEN
+		OnTouchListener onTouchListener = argsOnTouch.getValue();
+		tap(onTouchListener, eventTab, 6);
+
+		when(eventSwipe.getActionMasked()).thenReturn(MotionEvent.ACTION_POINTER_DOWN);
+		positionPointers(onTouchListener, eventSwipe, (float) 261.81, (float) 264.99);
+		when(eventSwipe.getActionMasked()).thenReturn(MotionEvent.ACTION_POINTER_UP);
```
jkuester, line 112:
```suggestion
```

This line is not needed right?

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.MockSettings;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(sdk=28)
+public class OpenSettingsDialogFragmentTest {
```
jkuester, line 30: Great tests! Nice to be checking all the different cases. We really need to upgrade to jUnit 5 soon so that we can leverage [parameterized tests](https://www.baeldung.com/parameterized-tests-junit-5) in some of these cases where the test code is basically exactly the same....

jkuester, line 30: Do you think we need any test cases around the `TIME_BETWEEN_TABS` logic?  Not sure I have ever done time manipulation in jUnit tests, so can't say how much of a pain that might be. (That being said, 500ms is not really that long for a `Thread.sleep` in the test case... so maybe we could just manipulate time the "normal" way :smile: )

jkuester (approved):
:rocket:

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
```diff
@@ -0,0 +1,83 @@
...
+import android.view.View.OnTouchListener;
+
+import androidx.annotation.Nullable;
+
+@SuppressLint("ValidFragment")
+public class OpenSettingsDialogFragment extends Fragment {
+
+	private final View view;
+	private int fingerTabCount = 0;
+	private long lastTimeTab = 0;
```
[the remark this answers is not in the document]
latin-panda: I completely mixed tab and tap 😵‍💫 in my head 
Fixed!

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
```diff
@@ -0,0 +1,83 @@
...
+	private GestureHandler swipeGesture;
+	private static final int TIME_BETWEEN_TABS = 500;
+
+	private final OnTouchListener swipeRightListener = new OnTouchListener() {
```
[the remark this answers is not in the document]
latin-panda: Done

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
latin-panda: Nice! I like functional

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
latin-panda: Done!

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
latin-panda: Nice! I've created [an issue](https://github.com/medic/cht-android/issues/260) to not forget.

latin-panda: This is nice, worked just fine. I think I'll keep using the `java.time.Clock` instead of the System's one... 
The funny thing is that `Clock myClock = Clock.fixed(Instant.ofEpochMilli(1), ZoneOffset.UTC);` has to be outside the `try` block otherwise there's a NPE when getting the millis with `java.time.Clock.systemUTC().millis()`.

latin-panda: I've added the test case covering the tap timeout :) Thanks!

latin-panda:
Feedback about unit tests is done, I'm moving this to AT 😃

README.md
```diff
@@ -15,6 +15,8 @@ Checkout the release notes in the [Changelog](CHANGELOG.md) page, our you can se
 
 Development guides are available in the "Android" section of the [Community Health Toolkit Docs Site](https://docs.communityhealthtoolkit.org/core/guides/android/). You will find instructions of how to setup your development environment, build and test new features, creates new branded apps, release, publish... and so on.
 
+## Settings page
+To open the app's Settings page, tap 5 times with one finger on the screen and then swipe right with two fingers, make sure to do this sequence fast.
```
mrjones-plip, line 19: @latin-panda - can you add another sentence here please? It's not clear to me why I would want to go to this settings page, so maybe just add something like:

> Once opened, you can change the URL of the CHT server, change the color of the background and change the session timeout.  Clicking "Save" will save the changes and close the settings page.

(I made all those up, of course!)

latin-panda: Done, thanks for the feedback

mrjones-plip (approved):

PR #256, opened 2022-04-04
Update pmd gradle config to exclude files and add support for xml

### LINKED ISSUE #255: PMD Processing errors for files that are not `.java`
**Describe the issue**
Currently, when generating PMD reports, the resulting HTML/XML files are `7.6 MB` in size!  :scream: This seems to be due to the fact that the reports include stacktraces for each of the "Processing errors" that were encountered (one for each file that PMD could not parse).  It appears that the PMD Gradle plugin by default only supports [parsing `.java` files](https://github.com/gradle/gradle/blob/master/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/PmdPlugin.java#L168) and since we are just trying to [parse everything in `src/`](https://github.com/medic/cht-android/blob/bd37e11d0fb335faf6391065ce23a720f4a8957f/build.gradle#L40) it is also trying to read all the images/xml/etc. 

**Describe the improvement you'd like**

PMD, itself. [supports parsing XML](https://pmd.github.io/latest/pmd_userdocs_cli_reference.html#supported-languages), but this has to be specially configured for the Gradle plugin. It would be nice to set that up and then exclude all the images and other files we do not want to parse.

A different approach would be to exclude all the `res` files from being parsed so that only `.java` are picked up as `source` for PMD, but this does not allow for analyzing the XML files.

jkuester:
Updates Gradle config for the PMD task so that `XML` files can be parsed.  Also excludes `png/mp3/bats` files since those also cannot be parsed by PMD.  

See https://github.com/gradle/gradle/issues/5451 for more info on why adding the `pmd-xml` dependency enables support for xml files...

Also, for the record here is the changes that happen to the dependency tree as a result of this change:

<table>
<tr>
	<td>master
	<td>255_pmd_config
<tr>
	<td>
<pre>
pmd - The PMD libraries to be used for this project.
\--- net.sourceforge.pmd:pmd-java:6.36.0
     +--- net.sourceforge.pmd:pmd-core:6.36.0
     |    +--- org.antlr:antlr4-runtime:4.7.2
     |    +--- com.beust:jcommander:1.48
     |    +--- commons-io:commons-io:2.6
     |    +--- net.sourceforge.saxon:saxon:9.1.0.8
     |    +--- org.apache.commons:commons-lang3:3.8.1
     |    +--- org.ow2.asm:asm:9.1
     |    \--- com.google.code.gson:gson:2.8.5
     +--- net.sourceforge.saxon:saxon:9.1.0.8
     +--- org.ow2.asm:asm:9.1
     +--- commons-io:commons-io:2.6
     \--- org.apache.commons:commons-lang3:3.8.1
</pre>
	</td>
	<td>
<pre>
pmd - The PMD libraries to be used for this project.
+--- net.sourceforge.pmd:pmd-java:6.36.0
|    +--- net.sourceforge.pmd:pmd-core:6.36.0
|    |    +--- org.antlr:antlr4-runtime:4.7.2
|    |    +--- com.beust:jcommander:1.48
|    |    +--- commons-io:commons-io:2.6
|    |    +--- net.sourceforge.saxon:saxon:9.1.0.8
|    |    +--- org.apache.commons:commons-lang3:3.8.1
|    |    +--- org.ow2.asm:asm:9.1
|    |    \--- com.google.code.gson:gson:2.8.5
|    +--- net.sourceforge.saxon:saxon:9.1.0.8
|    +--- org.ow2.asm:asm:9.1
|    +--- commons-io:commons-io:2.6
|    \--- org.apache.commons:commons-lang3:3.8.1
\--- net.sourceforge.pmd:pmd-xml:6.36.0
     +--- net.sourceforge.pmd:pmd-core:6.36.0 (*)
     +--- org.antlr:antlr4-runtime:4.7.2
     \--- commons-io:commons-io:2.6
</pre>
	</td>
</table>

Closes #255

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,13 @@ task pmd(type: Pmd) {
   ruleSets = []
   ignoreFailures = false
   source 'src'
+  exclude '**/**.png'
+  exclude '**/**.mp3'
+  exclude '**/**.bats'
+  dependencies {
+    pmd 'net.sourceforge.pmd:pmd-java:6.36.0'
+    pmd 'net.sourceforge.pmd:pmd-xml:6.36.0'
+  }
 }
 
 spotbugs {


latin-panda (approved):
Nice! Thanks for sending this

Issue #258, opened 2022-04-08
Make it clear when a training app is in use


abbyad:
**Is your feature request related to a problem? Please describe.**
From https://github.com/medic/cht-core/issues/6668: CHWs are often trained on one app which uses a test cht-core instance and then expected to use another app for live data entry on the production cht-core instance.  This can result in confusion as to which app to use, [most recently 50 CHWs entered 1800 documents in a test instance](https://github.com/medic/config-moh-mali/issues/98) that all should have been in the live instance.

**Describe the solution you'd like**
Make it obvious to users when they are using a test app, by showing a colored border in the app, and presenting a message when starting the app.

**Describe alternatives you've considered**
We had started with full explanations on detecting a training app, which included [looking for the URL in the About menu](https://github.com/medic/cht-docs/blob/69e5a724da479af83cadbc81e6524c754b8e344d/content/en/apps/guides/data/on-boarding-data-integrity.md#training-and-documentation). This could be handled CHT Core and controlled through settings, but that would require a much heavier lift. Managing it in custom builds provides a lightweight way to distinguish training apps.

**Additional context**
Making it clearer when a training app is being used serves towards the [user management at scale](https://github.com/medic/cht-roadmap/issues/58) initiative.
Here is a sample of what it could look like:
![image](https://user-images.githubusercontent.com/789512/162537366-1d784597-4639-4285-95b3-6d3b7e439519.png)

ngaruko:
Good to merge @latin-panda . [Commented](https://github.com/medic/cht-docs/pull/678#issuecomment-1103328428) on PR

PR #259, opened 2022-04-08
258 - Differentiate training app

abbyad:
Adds border and training app message

For https://github.com/medic/cht-android/issues/258

This will be useful for new deployments that want to differentiate their Training app, and would NOT affect any existing flavors.

[_**Update**:_ documentation in [cht-docs/678](https://github.com/medic/cht-docs/pull/678)]

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,6 +71,11 @@ jobs:
       with:
         lane: build
         options: '{ "flavor": "medicmobilegamma" }'
+    - name: Assemble gamma training
+      uses: maierj/fastlane-action@v1.4.0
+      with:
+        lane: build
+        options: '{ "flavor": "medicmobilegamma_training" }'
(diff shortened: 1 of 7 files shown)


latin-panda (changes requested):
Nice✨ Thank you for sending this! 

I think we need to add the flavor in [publish.yml](https://github.com/medic/cht-android/blob/master/.github/workflows/publish.yml#L69) too.

Do you think we should explain this flag in the docs? Maybe in this page: 
https://docs.communityhealthtoolkit.org/core/guides/android/branding/

abbyad:
Thanks @latin-panda! Good to know about the publish.yml, I'll do that. 

And yes, we should definitely document if we plan to include this in the app -- I'll start that now so that it is ready this week.

abbyad:
Documentation added through https://github.com/medic/cht-docs/pull/678.

Issue #257, opened 2022-04-05
Re-enable storage permissions and remove unbranded-test flavour


ngaruko:
[This ticket](https://github.com/medic/cht-android/issues/148) in cht-android re-enable the storage permission that was [removed last year](https://github.com/medic/cht-android/commit/9e0d1e28765c4c94206fc03b8db396f6c99b01f5), with a new flavour `unbranded-test` created.
I would [suggest](https://github.com/medic/cht-android/issues/148#issuecomment-1088399233) we remove it to avoid confusion in the future.
Also, think of re-enabling the storage permission for partners who use app launcher.
![image](https://user-images.githubusercontent.com/6979995/161831461-55aa87a0-9b70-48da-971c-93c347e452e4.png)

latin-panda:
Cross posting here for the record:
- `unbranded_test` is removed.
- Partners will need to add the permission by themselves, so they can control when to release it and when communicate the change to their CHWs. A partner might want to add a permission if they are using a feature that requires it. Because of that, we won't update any of the partners' flavor.

PR #261, opened 2022-04-13
257 - Remove unbranded test flavor

latin-panda:
# Description

This PR: 
- Removes the `unbranded_test` flavor. If we need to test a flavor without any of the permissions, then we can do that locally without adding a new flavor in the source code.

Note: Partners will need to add the permission by themselves, so they can control when to release it and when communicate the change to their CHWs. A partner might want to add a permission if they are using a feature that requires it. Because of that, we won't update any of the partners' flavor.

Ticket: https://github.com/medic/cht-android/issues/257

# License

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -192,11 +192,6 @@ jobs:
       with:
         lane: build
         options: '{ "flavor": "trippleeighty" }'
-    - name: Assemble unbranded_test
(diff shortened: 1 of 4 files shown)


latin-panda:
Hi @ngaruko can you please review this change? thanks!

ngaruko (approved):
Good job. Will avoid confusion going forward.

Issue #205, opened 2021-08-09
Changes in upcomming Android 12 may break the SMS capabilities


mrsarm:
In the upcoming Android 12 there is an API that changes the way can be used, and we use that feature in the code that handle the SMS capabilities.

Starting Android 11 the new linter from the Android Gradle Plugin warns about it (check [this](https://github.com/medic/cht-android/pull/202#discussion_r685426869)) so we needed to add an annotation in the source code to ignore the issue, but later we will need to remove the annotation and makes the necessary changes.

Warning when we remove the annotation:

![Screenshot from 2021-08-09 16-55-38](https://user-images.githubusercontent.com/1608415/128766381-2b69a49e-ac96-4e19-9ecc-2259b8b1d6a7.png)

mrsarm:
More info about the change: https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability

latin-panda:
- We need to properly upgrade the app to Android 12 (`compileSdkVersion 31`). The Google's target requirement hasn't been announced yet, but [looking at the trend](https://support.google.com/googleplay/android-developer/answer/9859152?visit_id=637824762632387275-2926058478&rd=1#targetsdk&zippy=%2Ctarget-api-level-requirements-for-play-console), it's probably going to be around 1st August 2022.

- We need to check for breaking changes like the one described by Mariano.

cc: @jonathanvq

craig-landry:
Temporarily making this "medium" priority as it pertains to the 1.0 release until we agree on if it must be included now or not.

jkuester:
This is ready for at on https://github.com/medic/cht-android/tree/205-Android_12.  You can see the PR for more details, but here is a breakdown of what I think needs to be tested:

1. Runs and drives on Android 12.
    1. Normal things work (can create/review/edit reports, etc)
1. SMS permissions:
    1. SMS: This requires more setup on the instance and then you have to create a report for your specific form. [Details](https://github.com/medic/cht-android/pull/262#issuecomment-1104443271)
    1. Expected behavior: when submitting a form that triggers and SMS message the following standard permission behavior should be followed:
        1. Display the prominent disclosure:
            1. ![image](https://user-images.githubusercontent.com/8441903/166579263-2fa52c0d-b93e-4a2d-82e1-77d9d7593c11.png)
[...]

PR #252, opened 2022-03-29
94 - Removing "_redirect" from app root url

latin-panda:
# Description 

This PR:

- Removes `DISABLE_APP_URL_VALIDATION` and therefore the App URL verification (AppUrlVerification class) isn't skipped anymore. 
- Stops appending `/medic/_design/medic/_rewrite/` to the App URL. 
- Adds some unit test coverage

Ticket: https://github.com/medic/cht-android/issues/94

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #262, opened 2022-04-13
205 - upgrade to Android 12

latin-panda:
# Description

This PR: 
- Upgrades to Android 12
- Fixes url validation for the Connexion Error page and added a test case to cover this validation. 
- Fixes deprecated attributes (backup strategy) in the manifest. 
- Adds `SEND_SMS` permission in the manifest, because when sending reports as SMS I was getting this error: `java.lang.SecurityException: Sending SMS message: uid 10655 does not have android.permission.SEND_SMS`
- Adds prominent disclosure when requesting `SEND_SMS` permission

Ticket: https://github.com/medic/cht-android/issues/205

# License


src/main/java/org/medicmobile/webapp/mobile/Alert.java
```diff
@@ -1,20 +1,42 @@
...
 import android.os.Vibrator;
+import android.os.VibratorManager;
 
 public class Alert {
-	private final MediaPlayer m;
-	private final Vibrator v;
+	private final MediaPlayer mediaPlayer;
+	private final Vibrator vibrator;
+
+	@SuppressWarnings("deprecation")
```
latin-panda, line 17: spotbugs lint still complains about the deprecated code even though we're already handling it with `Build.VERSION.SDK_INT >= XX` condition.

src/main/java/org/medicmobile/webapp/mobile/AppUrlVerifier.java
```diff
@@ -87,3 +91,22 @@ public static AppUrlVerification failure(String appUrl, int failure) {
...
+
+	public AppUrlVerificationTask(String appUrl) {
+		this.appUrl = appUrl;
+	}
+
+	@Override
+	public AppUrlVerification call() {
+		trace(this, "AppUrlVerificationTask :: Executing call, appUrl=%s", appUrl);
+
+		if (DEBUG && (appUrl == null || appUrl.isEmpty())) {
```
latin-panda, line 106: This code is from [here](https://github.com/medic/cht-android/blob/ebe095c6432d707e5b153c1526e4b82eded2f77a/src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java#L44).

src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -109,7 +80,23 @@ public void verifyAndSave(View view) {
...
-		new AppUrlVerificationTask(this).execute(appUrl);
+		TaskRunner taskRunner = new TaskRunner();
+		taskRunner.executeAsync(new AppUrlVerificationTask(appUrl), (result) -> {
+			trace(
+				this,
+				"SettingsDialogActivity :: Executing verification callback, result isOkay=%s, appUrl=%s",
+				result.isOk, result.appUrl
+			);
+
+			if (result.isOk) {
```
latin-panda, line 91: This is from [here](https://github.com/medic/cht-android/blob/ebe095c6432d707e5b153c1526e4b82eded2f77a/src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java#L48).

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
```diff
@@ -31,10 +41,17 @@ class SmsSender {
 
 	private final EmbeddedBrowserActivity parent;
 	private final SmsManager smsManager;
+	private Sms sms;
 
+	@SuppressWarnings("deprecation")
```
latin-panda, line 46: Spotbugs lint still complaining about deprecated code when we're already handling with `Build.VERSION.SDK_INT >= XX` condition

src/main/java/org/medicmobile/webapp/mobile/AppUrlVerifier.java
```diff
@@ -87,3 +90,22 @@ public static AppUrlVerification failure(String appUrl, int failure) {
...
+
+	public AppUrlVerificationTask(String appUrl) {
+		this.appUrl = appUrl;
+	}
+
+	@Override
+	public AppUrlVerification call() {
+		trace(this, "AppUrlVerificationTask :: Executing call, appUrl=%s", appUrl);
+
+		if (Utils.isDebug() && (appUrl == null || appUrl.isEmpty())) {
```
latin-panda, line 105: Not feeling strong about wrapping `BuildConfig.DEBUG` in `Utils.isDebug` for unit testing 🤷‍♀️

latin-panda:
Hi @jkuester can you please review? And thanks a lot for all the reviews and suggestions, I've learned tons! 🤩

src/main/java/org/medicmobile/webapp/mobile/AppUrlVerifier.java
latin-panda: MockStatic didn't work 🤔

jkuester (commented):
I have not finished here yet, but I have looked through all the SMS logic (not tests) and wanted to submit what I have so far.

Couple of questions I have after this initial look:
- Do I need to do anything special to get the app to ask for the SMS permission? I thought all I would need to do was send a message from the `Messages` tab, but that did not trigger it. (Maybe I need some special config on my CHT instance?)
- What is the actual problem we are trying to solve with the new URL code? (What was not working before?)

README.md
```diff
@@ -18,6 +18,9 @@ Development guides are available in the "Android" section of the [Community Heal
 ## Settings page
 To open the app's Settings page, tap 5 times with one finger on the screen and then swipe right with two fingers, make sure to do this sequence fast. Once opened, you can change the URL of the CHT server and click "Save" to save the changes and close the Settings page.
 
+## Settings Dialog
+To open the Settings Dialog page, tap 5 times with one finger on the screen and then swipe right with two fingers, make sure to do this sequence fast.
+
```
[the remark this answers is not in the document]
latin-panda: Oh thanks, I didn't noticed this bad merge

src/main/res/xml/backup_rules.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
latin-panda: Done

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
```diff
@@ -59,16 +76,52 @@ class SmsSender {
...
+		if (resultCode == RESULT_OK && this.sms != null) {
+			sendSmsMultipart(this.sms);
+			return;
+		}
+
+		trace(this.parent, "SmsSender :: Cannot send sms without Send SMS permission. Sms=%s", this.sms);
 	}
 
 //> PRIVATE HELPERS
+	private void setSms(Sms sms) {
```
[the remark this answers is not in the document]
latin-panda: Done

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
```diff
@@ -59,16 +76,52 @@ class SmsSender {
...
+		sendSmsMultipart(sms);
+	}
+
+	void resumeProcess(int resultCode) {
+		if (resultCode == RESULT_OK && this.sms != null) {
+			sendSmsMultipart(this.sms);
+			return;
+		}
+
+		trace(this.parent, "SmsSender :: Cannot send sms without Send SMS permission. Sms=%s", this.sms);
```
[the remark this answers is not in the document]
latin-panda: Done

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
```diff
@@ -281,4 +332,28 @@ private String getGenericFailureReason(Intent intent) {
 			}
 		}
 	}
+
+	static class Sms {
```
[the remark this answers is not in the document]
latin-panda: Thanks!

build.gradle
```diff
@@ -389,18 +389,18 @@ dependencies {
   // Latest version of androidx.core requires Android 12+
```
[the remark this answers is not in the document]
latin-panda: Looks like it's working, I just smoke test

src/main/AndroidManifest.xml
```diff
@@ -23,11 +23,16 @@
...
+  <uses-permission android:name="android.permission.SEND_SMS"/>
+
   <application android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
       android:allowBackup="false"
+      android:fullBackupContent="@xml/backup_rules_sdk_30_and_lower"
+      android:dataExtractionRules="@xml/backup_rules"
       android:largeHeap="true"
-      tools:ignore="GoogleAppIndexingWarning,LockedOrientationActivity,Instantiatable">
+      tools:ignore="GoogleAppIndexingWarning,Instantiatable,LockedOrientationActivity,UnusedAttribute">
```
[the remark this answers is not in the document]
latin-panda: The IDE will highlight the error, but also if you run `make lint`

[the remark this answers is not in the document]
latin-panda: Screen Shot 2022-04-20 at 1 27 31 pm

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
latin-panda: I first deleted the `build` folder, and run `make lint`

latin-panda: I quick searched in spotbugs config but haven't found anything yet 🤔

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
```diff
@@ -95,39 +148,37 @@ private String describe(Intent intent) {
...
+		return getBroadcast(parent, UNUSED_REQUEST_CODE, intent, flags);
 	}
 
 //> STATIC HELPERS
 	/**
-	 * @see https://developer.android.com/reference/android/telephony/SmsMessage.html#createFromPdu%28byte[],%20java.lang.String%29
+	 * @see <a href="https://developer.android.com/reference/android/telephony/SmsMessage#createFromPdu(byte[])">createFromPdu(byte[])</a>
 	 */
-	@SuppressLint("ObsoleteSdkInt")
+	@SuppressWarnings("deprecation")
```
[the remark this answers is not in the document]
latin-panda: I first deleted the build folder, and run `make lint`

latin-panda:
> Do I need to do anything special to get the app to ask for the SMS permission? I thought all I would need to do was send a message from the Messages tab, but that did not trigger it. (Maybe I need some special config on my CHT instance?)

This isn't from Message tab but from Reports tab(Enketo form), it took me awhile to _partially_ figure it out.
[When submitting a form](https://github.com/medic/cht-core/blob/6f3551a1b7b6d9efc5c9a507083f18960c70b9ed/webapp/src/ts/services/enketo.service.ts#L789), and if gateway is configured, then cht-core [will send the form as SMS](https://github.com/medic/cht-core/blob/6f3551a1b7b6d9efc5c9a507083f18960c70b9ed/webapp/src/ts/services/submit-form-by-sms.service.ts#L49).
I was in a rush to finish this before I leave, so I just made Enketo to always submit the form as sms in my local environment, instead of configuring the gateway and others things (that I still need to discover 😬)...

> What is the actual problem we are trying to solve with the new URL code? (What was not working before?)

Remove backslash at the end of the AppURL, this is a bug that we didn't catch while removing the `/medic/_design/medic/_rewrite/` from AppURL [in this PR](https://github.com/medic/cht-android/pull/252/files#diff-d9bfe28a0aee5b476da7e9b261a3636471bf81c0c1a205d6aafb12804abdfe8bL26). I discover the problem when testing the Connection Error Page and using Gamma Dev.

1. install the app
2. select gamma.dev (my connection with this server is slow)
3. immediately turn on airplane mode.
4. the app should show the [connection error page ](https://github.com/medic/cht-android/blob/master/src/main/res/layout/connection_error.xml)

Because Gamma Dev's url has a blackslash, `https://gamma.dev.medicmobile.org/` then it wasn't passing the validation and the Connection Error Page didn't display. 

I'll put this details in the PR description for QA.

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
latin-panda: Ah it works but it's deprecated: 
https://spotbugs.readthedocs.io/en/latest/annotations.html?highlight=suppresswarnings#edu-umd-cs-findbugs-annotations-suppresswarnings-deprecated

latin-panda: Funny, so I tried using `@SuppressFBWarnings` as they suggest but then the warning comes up again. With `@SuppressWarnings` the warning as gone. Then I noticed it's when it executes: `Task :compileUnbrandedDebugJavaWithJavac`

Screen Shot 2022-04-20 at 2 15 38 pm
Screen Shot 2022-04-20 at 2 16 04 pm

src/main/AndroidManifest.xml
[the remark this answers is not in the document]
jkuester: Ah got it, thanks!  

Unrelated, but do we still need to ignore `GoogleAppIndexingWarning` and `Instantiatable`? It seems like these may no longer need to be ignored...

src/main/java/org/medicmobile/webapp/mobile/SmsSender.java
jkuester: Okay, I think I see what I was missing. I was thinking that this was failing the build, but I think this is just the "normal" compiler warnings.  In that case, my vote would be to leave the warnings and not include the annotations. This is because we eventually want to update this code to not use the deprecated methods (after we raise our minSDK version). IMHO the only warnings we want to suppress are the ones where we know we do not want to change the way we are doing things (and so there is no reason to keep the warning around as a reminder...).  Ultimately, I am fine suppressing these if that makes more sense to you, though.

jkuester:
> This isn't from Message tab but from Reports tab(Enketo form), it took me awhile to partially figure it out. . . . (that I still need to discover :grimacing:)...

Thanks for the pointers here!  This makes so much more sense now.  For the record, I got really curious about this and dug in long enough to figure out how to get the SMS to send!  

- In `settings.js` doc, Set a phone number in the `settings` object for the name `gateway_number`
	- `"settings":{"gateway_number":"*the_phone_num*" ... }`
- In your form document (e.g. `form:enketo_widgets`), set `"xml2sms": true`

This will trigger SMS messages to be sent to the gateway number when submitting a form using the Android app. (Note that the gateway number cannot be the number of the phone that the app is running on (seems that you cannot sent SMS to yourself..)).

For more info: https://docs.communityhealthtoolkit.org/apps/guides/forms/app-form-sms/

src/main/java/org/medicmobile/webapp/mobile/AppUrlVerifier.java
jkuester: Wow! It appears that even the deep magic from `org.powermock.reflect.Whitebox.setInternalState` cannot touch the `BuildConfig` state....

Honestly, wrapping it feels like the right answer and your updates to the `Utils` class seem like the most straight-forward way to do this.  :+1: 

The only thing that bugs me is that mocking the static `Utils.isDebug` method is still annoying in the tests since you need the try-block and mockStatic stuff.  This is another great example of where a dependency injection framework would be helpful since we could just have an ordinary class (with non-static methods) that could act as our wrapper for `BuildConfig`. But instead of having to instantiate it everywhere or pass it into everything, the DI framework could just inject it for us.

src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java
```diff
@@ -27,7 +27,11 @@ public String getUrlToLoad(Uri url) {
 	}
 
 	public boolean isRootUrl(String url) {
-		return getAppUrl().equals(url);
+		if (url == null) {
+			return false;
+		}
+
+		return getAppUrl().equals(url.trim().replaceAll("/$", ""));
```
jkuester, line 34: Is this what is actually causing the bug around the "Connection Error Page"?  Is it just a race condition between hitting the `UrlHandler.onReceivedError` before the AppUrlVerifier task could save the cleaned URL?

Also, for the sake of code reusibility, does it make sense to make the `AppUrlVerifier.clean` method `public static`? Then we could use the same cleaning logic here...

jkuester (commented):
Okay, made it though the rest of the implementation code. Still need to look at the tests....

config/pmd.xml
```diff
@@ -11,7 +11,7 @@
     <exclude name="AvoidDuplicateLiterals"/>
     <exclude name="AvoidDuplicateLiterals"/>
     <exclude name="AvoidFieldNameMatchingMethodName"/>
-    <exclude name="MissingBreakInSwitch"/>
+    <exclude name="ImplicitSwitchFallThrough"/>
```
jkuester, line 14: Name changed in the new version of PMD

garethbowen (approved):
Awesome work @jkuester and @latin-panda !

Issue #207, opened 2021-08-18
Request "Approximate" location along to "Precise" location will be required in Android 12


mrsarm:
Android 12 will require apps to request "Approximate" location (`ACCESS_COARSE_LOCATION`) if they try to also access "Precise" location, check https://developer.android.com/about/versions/12/behavior-changes-12#approximate-location .

We have declared that permission in the app manifest, but at runtime we are only requesting the fine location. Not requesting the coarse granularity may cause Android to reject the access to the permission at runtime, and without any visual warning about it, only an error message in the device log is recorded (check https://developer.android.com/about/versions/12/approximate-location).

Also as mention in https://github.com/medic/cht-android/issues/189 , a refactor of how the location is requested is needed.

garethbowen:
As part of this we should track which users select which option (none, coarse, or fine), ie: show it on the About page, and store it in user telemetry. We could also go further and track all permissions that we request.

Ideally we would also include a way for users to change, so if they've picked coarse they can later launch the prompt again and change their permissions in the app.

Just my 2c... I'm not sure how hard this is, but it would be really useful to track in telemetry!

mrsarm:
Testing the App with Android 12 beta in an emulator, found the following results.

The geolocation didn't work well with the app as it is now (without changing the source code or the targeted SDK). The new UI that allows to select which precision of location is displayed as pointed out in the official documentation, and when selected fine precision, the geolocation data was collected as expected:

```json
  "geolocation": {
    "latitude": 37.421999,
    "longitude": -122.0839994,
    "altitude": null,
    "accuracy": 20,
    "altitudeAccuracy": null,
    "heading": null,
    "speed": null
  }-
```

But, if the user picks approximate location instead of precise location, it failed to collect the geolocation, without any notice to the user. This is what we get in the record:

```json
  "geolocation": {
    "code": -1,
    "message": "Geolocation timeout exceeded"
  },
```

This is bad because the user has the impression that it provided at least an aprox location but it didn't. The solution according with the documentation is to ask both permissions not just fine permission.

Targeting SDK 31 the problem is worst, because as it's pointed out in the documentation, the permission is rejected by Android, so after accepting the prominent disclosure the Android OS reject the permission automatically, and again the user has the impression that already accepted because in this case at least accepted the prominent disclosure.

[...]

PR #263, opened 2022-04-18
Handle Coarse Location permissions in Android 12

latin-panda:
**Note that this PR is pointing to [205-Android_12](https://github.com/medic/cht-android/tree/205-Android_12) branch. Squash/merge this PR AFTER the [Upgrade Android 12 PR is merged](https://github.com/medic/cht-android/pull/262).**

# Description 

This PR:
1. If fine (precise) or coarse (approximate) permission is granted then the user can continue filling out the form
2. If just approximate permission is granted, then we'll ask for the permission again and see if we can get precise location.
3. if none of those location permissions are granted, then we'll ask for those permissions.

Screenshot when #2 point happens: 

Closes https://github.com/medic/cht-android/issues/207

# License


latin-panda:
Hi @jkuester can you please review this small one? Thanks!

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -254,17 +254,11 @@ public boolean getLocationPermissions() {
...
 		}
 
-		if (settings.hasUserDeniedGeolocation()) {
-			trace(this, "getLocationPermissions() :: user has previously denied to share location");
-			locationRequestResolved();
-			return false;
-		}
-
-		trace(this, "getLocationPermissions() :: location not granted before, requesting access...");
+		trace(this, "getLocationPermissions() :: Fine or Coarse location not granted before, requesting access...");
```
latin-panda, line 261: If we're granted with just approximate then let the user continue using the form, next time (this code) try again to get precise location.



Open to suggestions

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -266,17 +266,11 @@ public boolean getLocationPermissions() {
 		boolean hasCoarseLocation = ContextCompat.checkSelfPermission(this, ACCESS_COARSE_LOCATION) == PERMISSION_GRANTED;
 
 		if (hasFineLocation && hasCoarseLocation) {
-			trace(this, "getLocationPermissions() :: already granted");
+			trace(this, "getLocationPermissions() :: Fine and Coarse location already granted");
 			return true;
 		}
 
-		if (settings.hasUserDeniedGeolocation()) {
```
jkuester, line 273: Does this mean we are never going to stop asking the user for location permissions (even if they have explicitly denied them)?

jkuester: Oh wait, IIRC Android will stop asking after the user has denied (2 times?) right?  But would we still be popping the prominent disclosure every time?

src/main/java/org/medicmobile/webapp/mobile/RequestLocationPermissionActivity.java
```diff
@@ -34,10 +36,20 @@ public class RequestLocationPermissionActivity extends FragmentActivity {
...
-			boolean allGranted = grantedMap.size() > 0 && !grantedMap.containsValue(false);
-
-			if (allGranted) {
-				trace(this, "RequestLocationPermissionActivity :: User allowed location permission.");
+			boolean fineLocationGranted = Optional
+				.ofNullable(grantedMap.get(ACCESS_FINE_LOCATION))
+				.orElse(false);
+			boolean coarseLocationGranted = Optional
+				.ofNullable(grantedMap.get(ACCESS_COARSE_LOCATION))
+				.orElse(false);
```
jkuester, line 44:
```suggestion
			grantedMap.getOrDefault(ACCESS_FINE_LOCATION, false);
			grantedMap.getOrDefault(ACCESS_COARSE_LOCATION, false);
```

Love the Optional use here, but the built-in methods on Map are probably more convenient...

jkuester (changes requested):
Looks good!

Would be nice to have a couple unit tests specifically for testing the behavior when one of the location permissions is granted but not the other.  

Specifically, I am thinking that in the EmbeddedBrowserActivityTest it would be nice to verify that when the user grants one, the activity continues, but the next time the user is prompted again for the other location permission.

Then, in the RequestLocationPermissionActivityTest it would be nice to just cover the case were the user only grants one of the permissions (both from the normal prompt and from the settings) and assert the expected behavior occurs...)

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
latin-panda: It will behave the same as Storage and Send SMS permissions, if the user hasn't granted the permissions, then we try again asking them in case the have change their mind. Ideally we would like to have the GPS anyways for the forms reports. 
If they have denied too many times or have selected "Never ask again", we still ask them to grant the permission and redirect them to the app's settings page so they grant it manually because at the point we can't do it automatically anymore. 

The difference here is that now we have approximate and accurate location, I'm going to with asking for accurate location even if they just granted approximate, it won't block them to use the forms, they can keep choosing "approximate" 🤔

src/main/java/org/medicmobile/webapp/mobile/RequestLocationPermissionActivity.java
jkuester: Nope, you were right!  Looking at this again, you _could_ get an NPE since the map might contain a `null` value and Java will throw an exception if you try and assign `null` to a primitive `boolean`.  This is good like you have it!

jkuester (approved):
This should be good to go now!

Issue #266, opened 2022-04-23
Border not showing red for production URL in dev app


abbyad:
**Describe the bug**
When using production URL in dev app the border of the app is meant to be red, but shows blue or grey.

**To Reproduce**
1. Install unbranded app
1. Use cht.app.medicmobile.org as the url
1. See border color not being red

**Expected behavior**
When using a *.app.medicmobile.org URL in the unbranded app I expect the border to appear, and be red.

**Screenshots**
![Screenshot_20220421-095806-400](https://user-images.githubusercontent.com/789512/164836132-6db7a241-cc8b-4ccd-b7e1-841591dd26f8.jpg)

**Environment**
- Android Version: 11
- App Version: v0.11.0-webview

**Additional context**
[This code](https://github.com/medic/cht-android/blob/5218167a3ea3c7b66767fa8296c405ec5c2e9608/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java#L87-L93) adds the border but the color set with `setBackgroundColor` which uses the int of the resource itself, not the color it is defining -- which essentially means the border color is random since the resource ID can change with different builds. With #258 the color is set correctly [by using `setBackgroundResource`](https://github.com/medic/cht-android/blob/5218167a3ea3c7b66767fa8296c405ec5c2e9608/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java#L99).

latin-panda:
Ready for AT in this [PR](https://github.com/medic/cht-android/pull/267)

lorerod:
**Phone:** Galaxy A01 Core
**Android Version:** 10
**Android System WebView Version:** 97.0.4692.98

### Reproducible on Master
**Platform:** Mobile App. APK generated with branch `master`
<details>
<summary> Environment: brac-ug.app.medicmobile.org </summary>

- The border is not red

![brag-upOnMaster](https://user-images.githubusercontent.com/21312057/166493038-f2b0f3a5-37cb-4ae8-8a84-e29c9cd58f9c.jpg)

![brag-ugOnMasterLogin](https://user-images.githubusercontent.com/21312057/166493034-b9be46e5-1684-417f-8ef6-53da595f2901.jpg)

</details>

<details>
<summary> Environment: cht.app.medicmobile.org </summary>

- The border is not red

[...]

PR #267, opened 2022-04-23
266 - Set background color as specified in resource

abbyad:
For #266


latin-panda (approved):
Looks good, but I'm not understanding the feature completely. 

@abbyad, We're now highlighting the training apps with red border colour, why is it important to highlight apps using `*.app.medicmobile.org` in unbranded flavor? What's the history? 🙂

Thanks!

abbyad:
We wanted Medic-hosted deployments to use the builds (flavors) we were making for them, rather than the unbranded app, which was used frequently for testing or training. Having the border helped avoid mistakes, or alert trainers.

abbyad:
Thanks for the review @latin-panda, and I'll leave it to you to move the issue to the right milestone and merge when appropriate.

latin-panda:
@abbyad, do you think it makes sense to add a toast explaining why is red? Similar to when it's training app? 
"You're using Medic's instance" 
Sorry I just kept thinking on this one 😅

abbyad:
That's a good idea. A flavor can choose to have no toast by setting the message to an empty string.

PR #269, opened 2022-05-09
Remove duplicate switch case that was accidentally added

jkuester:
Thanks to some poor git merging magic, we ended up with a duplicate `switch` case that needs to be removed.


mrjones-plip (approved):
these are the same 3 lines as above - lgtm

PR #272, opened 2022-05-24
Add moh_mali_chw brand for APK

ernestoteo:
[#220](https://github.com/medic/config-moh-mali-chw/issues/220)


latin-panda:
Hi @binokaryg and @ernestoteo just in case the heads up, new flavors should have branched off from the [last stable release](https://github.com/medic/cht-android/releases/tag/v0.11.1). [See docs](https://docs.communityhealthtoolkit.org/core/guides/android/releasing/#new-flavor-release) 
The current state of master correspond to the upcoming version 1.0.0 (not released yet) that has important breaking changes like dropping support to xwalk (android 4)

latin-panda (changes requested):
In addition to my previous comment. I noticed [that step # 4 ](https://docs.communityhealthtoolkit.org/core/guides/android/branding/#2-new-brand)is missing :)

PR #274, opened 2022-05-24
220 moh mali chw

ernestoteo:
[#220](https://github.com/medic/config-moh-mali-chw/issues/220)


binokaryg:
When are we expecting to get the logo and app name?

mrjones-plip:
We should be sure to use the [new training app features](https://docs.communityhealthtoolkit.org/apps/guides/onboarding/#setting-up-a-training-app) in 1.0.0!

PR #275, opened 2022-05-25
commenting SEND_SMS permission

latin-panda:
# Description

Projects should add this permission if they want to send forms as sms. 
When publishing in Play Store, projects needs to justify why they need the permission

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -19,11 +19,12 @@
   <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT"/>
   <uses-permission android:name="com.android.launcher.permission.UNINSTALL_SHORTCUT"/>
 
-  <!-- READ_EXTERNAL_STORAGE required if users want to include photos from their phone  -->
+  <!-- READ_EXTERNAL_STORAGE is required if users want to include photos from their phone  -->
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
 
-  <!-- SEND_SMS required if users want to send reports as SMS -->
+  <!-- SEND_SMS is required if users want to send reports as SMS.To able the feature add the following line:
   <uses-permission android:name="android.permission.SEND_SMS"/>
+  -->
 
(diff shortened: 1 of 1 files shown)


src/main/AndroidManifest.xml
```diff
@@ -19,11 +19,12 @@
   <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT"/>
   <uses-permission android:name="com.android.launcher.permission.UNINSTALL_SHORTCUT"/>
 
-  <!-- READ_EXTERNAL_STORAGE required if users want to include photos from their phone  -->
+  <!-- READ_EXTERNAL_STORAGE is required if users want to include photos from their phone  -->
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
 
-  <!-- SEND_SMS required if users want to send reports as SMS -->
+  <!-- SEND_SMS is required if users want to send reports as SMS.To able the feature add the following line:
   <uses-permission android:name="android.permission.SEND_SMS"/>
```
jkuester, line 26: What does our SMS code do now, with this commented out, if it still gets triggered?  Just wondering if we crash hard, or silently ignore the SMS, or do we log some kind of error?

latin-panda: CHT-Core checks if the [sms is available ](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/submit-form-by-sms.service.ts#L21) and logs if it's not available
Even if it goes ahead and tries to send the sms, we have still [check for the permission and handle appropriately if the permission is granted or deny](https://github.com/medic/cht-android/blob/master/src/main/java/org/medicmobile/webapp/mobile/RequestSendSmsPermissionActivity.java), it logs a message either way and doesn't crash the app. 
Earlier when I tested this and bypass the cht-core validation, I get the prominent disclosure, then I click allow, but since the app couldn't grant the permission because it's not in the manifest, I'm redirected to the app's system settings page, from there I don't have option to enable the permission (because it's not predefined in the manifest)

jkuester (approved):

PR #279, opened 2022-06-02
V0.11.0 moh niger chw

ernestoteo:
[#51](https://github.com/medic/config-moh-niger/issues/51)


latin-panda:
Hi @ernestoteo I noticed the branch name is [v0.11.0-moh-niger-chw](https://github.com/medic/cht-android/tree/v0.11.0-moh-niger-chw) but the PR's base is `v1.0.x`. 
Maybe the PR's base should be `v0.11.x`?

Issue #282, opened 2022-06-09
Fix Implicit Intent Vulnerability 


latin-panda:
**Describe the bug**
We got the following error in Google Play Store:

![Screen Shot 2022-06-09 at 2 07 55 pm](https://user-images.githubusercontent.com/66472237/172786300-ba769278-318c-470a-8be6-c24d1c2e4e2a.png)

These links give more info about the error and fixes:
- [Remediation of Implicit Internal Intent Vulnerability](https://support.google.com/faqs/answer/10399926)
- [CWE-927: Use of Implicit Intent for Sensitive Communication](https://cwe.mitre.org/data/definitions/927.html)
- [Security tips: Use of intents](https://developer.android.com/training/articles/security-tips.html#use-intents)

>Applications that send broadcasts without specifying the broadcast target may have
these broadcasts intercepted by malicious apps on the same device.

(source: [Android_app_vulnerability_classes.pdf](https://github.com/medic/cht-android/files/8867752/Android_app_vulnerability_classes.pdf))

**Additional context**
Possibly introduced in [this PR](https://github.com/medic/cht-android/pull/165).

latin-panda:
When testing this issue, make sure these features are working fine:
- https://github.com/medic/cht-android/issues/163
- https://github.com/medic/cht-android/issues/148

latin-panda:
This is ready for AT [in this PR](https://github.com/medic/cht-android/pull/283), please consider to test these issues[ mentioned here](https://github.com/medic/cht-android/issues/282#issuecomment-1151875152).

latin-panda:
We can't replicate, in our local, the same "Implicit Intent Vulnerability" error from Play Store. The lint and the static check tools aren't detecting it, it's a check from Play Store.  Once we ensure this fix didn't affect badly the existing functionality, then we're going to release and deploy to Play Store, wait until it finishes the checks and see if the error appears again.

tatilepizs:
**Config:** Covid-19
**Environment:** Local with Docker helper script
**cht-core version:** master
Platform: Mobile App. APK generated with branch `282-Fix_Implicit_Intent_Vulnerability`

### Test - Ticket #163 

**Test scenario:**
Tested with four different phones. Removed internet connection to see the page indicating the error, after it was connected again to the internet, click on `Retry`, and the login page was displayed successfully.

<details>
<summary>Android 5.1</summary>

**Phone:** Alcatel OneTouch 50560
**Android System WebView Version:** 95.0.4638.74

[...]

Issue #163, opened 2021-03-15
Improve connection errors UX, especially for the XWalk-Webview migration process


mrsarm:
Quote from ticket #134 :

> Firstly, the service worker cache is not migrated, so if the user upgrades and the loses internet an ugly error message is shown.
> ![image](https://user-images.githubusercontent.com/10531066/99542121-aa51b900-297f-11eb-8b40-24f0ebba7183.png)

This ticket is not about implementing the service worker cache migration but improve the connection errors UX. Is not just the errors displayed have a bad look & feel, but also during the migration process it doesn't display a "Retry" button, like in other cases to retry after recover the connection. This cause the user to loss the app data during the migration, which is a big issue as stated in the ticket.

So, there are 2 things to improve:
1. When there are errors in the connection show an UI like the one proposed by @n-orlowski in the migration ticket:
   ![workflow](https://user-images.githubusercontent.com/59581790/102176167-de30d900-3e6e-11eb-849f-0c4ed09ff2a1.png)
2. Always ensure that the user can retry after the connection fails, not just when trying to migrate the app, eg. when the user install the app for the fist time sideloading the installer without internet connection.

### Important consideration

The workflow proposed has a special view for timeout errors that says: _"We are having trouble migrating your data. **You can wait a bit longer or cancel the upgrade**"_. And there is a "Cancel" button that:

> If user cancels, they are taken to the log in screen.

I'm not sure whether the user will understand what "Cancel" means on this context, and whether we should provide this option to the user.

The user may think that "Cancel" means cancel the upgrade process of the app, not the data, so cancel in the understanding of a normal user could be: _"I'm cancelling the installation of this new version of the app, so the old version will be kept along my data"_.

Also we should add in that same view a "Retry" button as in the other views to give the user the ability to retry.

CC @garethbowen @craig-landry @kennsippell to discuss whether a "Cancel" option is a good idea.

Then if it is OK, I can continue the communication of the action with @n-orlowski to find the best way of letting the users know what they are opting in.

garethbowen:
I think the cancel button was added when we were thinking the migration may get stuck in an infinite restart loop. The solution you've found for the migration means this is no longer the case, so it's not relevant. If the user does not have a service worker cache, nor an internet connection, then it doesn't make sense to "cancel" because we can't show them the login page anyway.

[...]

PR #283, opened 2022-06-09
282 - Fix Implicit Intent Vulnerability

latin-panda:
# Description
This PR:
- Adds package name to make intents explicit.

Ticket: https://github.com/medic/cht-android/issues/282

# License

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/ConnectionErrorActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/ConnectionErrorActivity.java
@@ -53,7 +53,9 @@ public class ConnectionErrorActivity extends ClosableAppActivity {
 	// Retry connection
 	public void onClickRetry(View view) {
 		trace(this, "onClickRetry()");
-		sendBroadcast(new Intent("retryConnection"));
+		Intent intent = new Intent("retryConnection");
+		intent.setPackage(this.getPackageName());
+		sendBroadcast(intent);
 		showSpinnerLayout();
 	}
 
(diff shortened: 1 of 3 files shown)


latin-panda:
@garethbowen In theory, it should be enough by just setting the package name, the intent will only be accessible inside the package and fixes the concerns in the "vulnerability" criteria. I found another 2 intents that could be considered as implicit, I fixed those as well, I tested that they still work fine. 

I couldn't find a lint option to catch this, so I opened a question in stackoverflow.  

1. Should this be ATed?
2. Is this going to be released as production or beta in Play Store?
3. Would this be a v1.0.1?
4. Do we want to wait for [this work ](https://github.com/medic/cht-android/pull/281) and release everything together? 

Thanks

garethbowen:
@latin-panda Have you managed to verify that this change still functions the way we expect? It was added as part of https://github.com/medic/cht-android/issues/163 so it should be possible to try and recreate the issue.

> Should this be ATed?

Yes, to make sure it hasn't regressed.

> Is this going to be released as production or beta in Play Store?

I think now that we have the production track going we should keep it going.

> Would this be a v1.0.1?

I think that'd be good, yes. We've fixed a bug without changing functionality so a service pack bump is appropriate.

> Do we want to wait for https://github.com/medic/cht-android/pull/281 and release everything together?

If they line up perfectly that's fine, but otherwise each should be released as soon as possible.

latin-panda:
@garethbowen Yes, I've tested the areas that changed and they are working find. 
Once approved this is approved, I'll move the ticket to AT and notify QA for second pair of eyes, before starting the minor release.

latin-panda:
@garethbowen since this is a patch release, do we want to still make release notes? Thanks

garethbowen:
@latin-panda Yes, it's best to update the changelog for every release.

garethbowen (approved):
Nice one!

PR #286, opened 2022-06-14
change order of checks for adding the border color

njogz:


src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -84,19 +84,19 @@ public void onReceiveValue(String result) {
...
-		// Add a noticeable border to easily identify a training app
-		if (BuildConfig.IS_TRAINING_APP) {
+		// Add an alarming red border if using configurable (i.e. dev)
+		// app with a medic production server.
+		if (settings.allowsConfiguration() && appUrl != null && appUrl.contains("dev.medicmobile.org")) {
 			View webviewContainer = findViewById(R.id.lytWebView);
 			webviewContainer.setPadding(10, 10, 10, 10);
-			webviewContainer.setBackgroundResource(R.drawable.training_background);
+			webviewContainer.setBackgroundResource(R.drawable.warning_background);
 		}
```
garethbowen, line 100: While you're working on it, let's improve the code a little too.

1. There's no point setting the background resource twice, so let's use if/else.
2. The production check is hard to read and should be named so a casual observer can understand what's going on. We could either pull out the production URL as a constant string and name it well, or (as in my example) pull out a method to check `isProduction`. This makes also makes it less likely for bugs like switching to "dev" to sneak through, because it'll be obvious from the method name that it's doing the wrong thing!
3. There's also duplicate code - we should pull out a private method to set the background.

For example...

```
private void setBackground(Object background) { // fix the type here...
	View webviewContainer = findViewById(R.id.lytWebView);
	webviewContainer.setPadding(10, 10, 10, 10);
	webviewContainer.setBackgroundResource(background);
}

private boolean isProduction(String appUrl) {
	return appUrl != null && appUrl.contains("app.medicmobile.org");
}

...

if (settings.allowsConfiguration() && isProduction(appUrl)) {
	setBackground(R.drawable.warning_background);
} else if (BuildConfig.IS_TRAINING_APP) {
	setBackground(R.drawable.training_background);
}
```

I haven't tested any of this so please treat it as pseudo code but to me this is much more readable and maintainable.

garethbowen (changes requested):
Looks good.

One little bug (I think?) and a suggested refactor for readability.

Issue #287, opened 2022-06-14
Can not install two flavors of CHT Android app of same version. 


yrimal:
**Describe the bug**
Can not install two different apps of CHT Android or it's flavors. For example, if I am a supervisor for two of the medic's deployments and both have Android app of same version, I can't install one of the app if another is already deployed. This happens with CHT Android app as well. 

**To Reproduce**
1. Download and install CHT App for [CHIS](https://github.com/medic/cht-android/releases/download/v1.0.0/cht-android-v1.0.0-chis_ne-arm64-v8a-release.apk)
2. Download and try installing [Unbranded app](https://github.com/medic/cht-android/releases/download/v1.0.0/cht-android-v1.0.0-unbranded-arm64-v8a-release.apk) or any [Muso Mali](https://github.com/medic/cht-android/releases/download/v1.0.0/cht-android-v1.0.0-musomali-arm64-v8a-release.apk) or any other apk of same version. 
3. You will see an error saying app wasn't installed. 
4. If you'd tried Step 2 using `adb`, you'd get error like below: 
```
adb: failed to install /Users/yuvraj/Downloads/cht-android-v1.0.0-musomali-arm64-v8a-release.apk: Failure 
[INSTALL_FAILED_CONFLICTING_PROVIDER: Scanning Failed.: Can't install because provider name 
org.medicmobile.webapp.mobile.fileprovider (in package org.medicmobile.webapp.mobile.musomali) 
is already used by org.medicmobile.webapp.mobile.chis_ne]
```

**Expected behavior**
Second app should be installed without error. 

**Logs**
Adb Log: 
```
adb: failed to install /Users/yuvraj/Downloads/cht-android-v1.0.0-musomali-arm64-v8a-release.apk: Failure 
[INSTALL_FAILED_CONFLICTING_PROVIDER: Scanning Failed.: Can't install because provider name 
org.medicmobile.webapp.mobile.fileprovider (in package org.medicmobile.webapp.mobile.musomali) 
is already used by org.medicmobile.webapp.mobile.chis_ne]
```

**Environment**
- Android Version: (eg: 11) 
- App Version: (eg: 1.0)

**Additional context**
Different version of CHT Android app (for example: CHIS v1.0 and Muso mali v0.9) can be installed side by side.

latin-panda:
Some notes:
- We set the file provider in the [manifest](https://github.com/medic/cht-android/blob/master/src/main/AndroidManifest.xml#L74)
[...]

PR #284, opened 2022-06-10
add change log for 1.1.0 release

njogz:

PR #288, opened 2022-06-15
287 - Allow installing multiple flavors in the same device

latin-panda:
# Description

This PR
- Adds `applicationId` to `unbranded` flavor, so every flavor keeps defining this property and ensure there aren't conflicts by having default values. Consistency. 
- Uses `applicationId` to define provider in manifest.

Ticket: https://github.com/medic/cht-android/issues/287

# License

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -198,6 +198,7 @@ android {
       // src/main
       dimension = 'brand'
       buildConfigField "boolean", "IS_TRAINING_APP", 'true'
+      applicationId = 'org.medicmobile.webapp.mobile'
     }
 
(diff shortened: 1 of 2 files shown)


latin-panda:
I have tested installing several APKs in the same devices Android 11, I have also tested that I can log in and use the apps normally while being in my device at the same time. I also tested that the original feature, File Picker, is still working the same as before. 

@garethbowen and/or @jkuester pls review, we'll try to release it with everything else in [v1.1.0](https://github.com/medic/cht-android/pull/284) 
FYI @njogz

PR #289, opened 2022-06-16
remove IS_TRAINING_APP setting from unbranded app

njogz:


latin-panda (approved):
LGTM! Thank you for the fast PR!

PR #295, opened 2022-07-28
Update Moh Mali CHW logo

ernestoteo:
[#362](https://github.com/medic/config-moh-mali-chw/issues/362)


binokaryg (commented):
The image looks blurry.
I found [this one](https://30minutes.net/wp-content/uploads/2019/05/gouvernement.jpg) on the web. Can we use it instead? It might need some cropping though.

binokaryg (commented):
Why is the area covered by the new image is smaller that the previous one?
image

binokaryg (approved):
Looks alright.

PR #301, opened 2022-11-23
Generate new flavor for MOH Mali CHW training based on CHT-4.0

ernestoteo:


binokaryg (approved):
Will it be a problem that the app logo is same as the earlier training app?
https://github.com/medic/cht-android/blob/moh_mali_training_app/src/moh_mali_chw_training/res/mipmap-xxxhdpi/ic_launcher.png

ernestoteo:
> Will it be a problem that the app logo is same as the earlier training app? https://github.com/medic/cht-android/blob/moh_mali_training_app/src/moh_mali_chw_training/res/mipmap-xxxhdpi/ic_launcher.png

No, seeing that it is the same application and it is intended for training just like the first one.
The only difference is that this one runs on CHT 4.0

Issue #302, opened 2023-02-14
New branded app for ICM: Flourish


binokaryg:
Ticket to keep track of the steps needed to release the new branded app "Flourish" for ICM.
APK will be shared to the partner so that it can be sideloaded.
Publishing to the Google Play Store will be done by the partner with our support later, if necessary.
The previously built [app](https://play.google.com/store/apps/details?id=org.medicmobile.webapp.mobile.icm_ph_chc) is already on the Play Store under partner's own account.

URL: http://flourish.caremin.com/
Logo: same as [before](https://github.com/medic/cht-android/blob/master/src/icm_ph_chc/res/mipmap-xxxhdpi/ic_launcher.png)
Name: `Flourish`

### Tasks
- [x]  Add PR
- [x]  Merge PR
- [x]  Add signing keys
- [x]  Release cht-android with the new flavor
- [x]  Share APK to the partner

binokaryg:
All tasks completed. APK and AAB files have been shared with the partner.

PR #303, opened 2023-02-15
Adding files for new flavor icm_flourish

binokaryg:
for #302


ernestoteo (approved):
Good @binokaryg ,
It looks good to me.

Issue #310, opened 2023-05-23
Add ability for branded APKs to deep link/register intents to wildcard domains


mrjones-plip:
**What feature do you want to improve?**
Right now we tightly bind the APKs intent to the `app_host` (see [step 3 in "New Brand"](https://docs.communityhealthtoolkit.org/apps/guides/android/branding/#2-new-brand)). This does not enable a way to link to something like `*.cht.domain.com`.  

**Describe the improvement you'd like**
Enable an extensible way to add a wildcard domain for all intents.  

**Describe alternatives you've considered**
NA

**Additional context**
This will work well when we deliver on #309.  We also already [have a deployment that needs this](https://github.com/medic/cht-android/releases/tag/untagged-fb666090af5fac8eb5b5).

jkuester:
@mrjones-plip @kennsippell I am slightly concerned that wildcards in the intent might be more trouble than it is work in the long term (e.g. when we go to fix #308).  Auto-verifying a wildcard intent in Android 12+ [is going to _require_](https://developer.android.com/training/app-links/verify-android-applinks#multi-subdomain) the `assetlinks.json` file be published _"at the root hostname"._  What is tricky here is that in most of these cases I can think of where the wildcard intent would be useful, there will probably not actually be a CHT instance running on the root domain. So, the deployment of that file would have to be a custom/bespoke process for each case....

The Android docs I linked note that registering _multiple links for the intent_ works different from a wildcard intent.  When multiple links are registered, the `assetlinks.json` file needs to be accessible at each of the links.  Technically this will be more moving parts when managing multiple CHT instances because you will end up with `n` files that have to be deployed.  However, a likely solution to #308 is to add support to cht-core/cht-conf for just deploying the file to the CHT server along with the rest of the configuration....

Also, for what it is worth, additional custom links can already be registered by specific brands just by creating a custom `AndroidManifest.xml` file for the brand and adding something like this:

```xml
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
  xmlns:tools="http://schemas.android.com/tools"
  package="org.medicmobile.webapp.mobile">

  <application>
    <activity android:name="AppUrlIntentActivity"
      android:launchMode="singleInstance"
      android:exported="true">
      <intent-filter>
        <action android:name="android.intent.action.VIEW" />
        <category android:name="android.intent.category.DEFAULT" />
        <category android:name="android.intent.category.BROWSABLE" />
[...]

PR #311, opened 2023-05-24
Add one intent-filter for each county in moh_kenya_echis flavor

kennsippell:
#310 

~https://github.com/medic/cht-android/pull/306/commits/b2e815fe383655156871a702980a4ca7a731ef8a we used a new resource string which specified the host (including optional wildcard) to be used by the intent.  We can't do exactly what we did in #307 because it will require a change to existing flavors. This slight variation creates two data elements for the same intent-filter. It's a bit concerning because, we're registering the following:~

Scenario | app_host | intent_host | Behavior
-- | -- | -- | --
Branded Normal (eg. DTree) | dtree.com | empty | Good
eCHIS Kenya | empty | *-echis.health.go.ke | Good
Unbranded | empty | empty | No intents work

~I welcome input to confirm this is the best path available.~

~I can't find any way to: add conditions in intent-filters, use `app_host | intent_host`, or dynamically add intents in code.The only other viable options I've found is to achieve this by overwriting AndroidManifest.xml directly for flavors which want this functionality.  To achieve this, I believe we'd need to duplicate the list of urls (cannot reuse the data in instances.xml). I suspect it's not very pretty. Please let me know if there is a better way to do this!~

Uses AndroidManifest.xml to add intent-filters per county. moh_kenya_echis only.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -64,7 +64,6 @@
         <category android:name="android.intent.category.DEFAULT" />
         <category android:name="android.intent.category.BROWSABLE" />
         <data android:scheme="@string/scheme" android:host="@string/app_host" android:pathPattern=".*"/>
-        <data android:scheme="@string/scheme" android:host="@string/intent_host" android:pathPattern=".*"/>
       </intent-filter>
     </activity>
     <activity android:name="UpgradingActivity"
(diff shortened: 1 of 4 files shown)


jkuester (approved):
Can confirm that all URLs are properly registered:
image

LGTM!

kennsippell:
Thanks @jkuester .  Do you think we should merge this into #313 or keep them separate?

jkuester:
@kennsippell I say just merge it now and they can both go into master together on #313!

Issue #309, opened 2023-05-23
Improve ability to have large list of instances to choose from


mrjones-plip:
**What feature do you want to improve?**
Right now there's a way to select a CHT Instance on the "unbranded" [play store app](https://play.google.com/store/apps/details?id=org.medicmobile.webapp.mobile).  It is limited though and can not be extended by normal [branding techniques](https://docs.communityhealthtoolkit.org/apps/guides/android/branding/): 

![android select](https://github.com/medic/cht-android/assets/8253488/d9f45d97-1ed2-4033-b7d5-6bff07019646)

**Describe the improvement you'd like**
For large deployments that have many CHT instances, like a national deployment, it would good if we could extend this to have:
* having an external XML file that any APK can use as part of their branding to define a list of CHT instances to show pre-login
* have a way to search long lists of CHT instances per above
* show confirmation dialog when making selection

**Describe alternatives you've considered**
NA

**Additional context**
We have an active deployment that is already using this [with an alpha APK](https://github.com/medic/cht-android/releases/tag/untagged-fb666090af5fac8eb5b5)

mrjones-plip:
accidentally closed with accidental merge in #306.  Re-opening for work being done in #313

PR #313, opened 2023-05-24
Add Flavor for MoH Kenya eCHIS

kennsippell:
Recreating #306 after erroneous merge and revert.

* Creates new flavor for moh_kenya_echis (config, resources, keystore, git actions)
* Loads a list of URLs from an XML resource (Fixes https://github.com/medic/cht-android/issues/309)
* Hides the Custom option based on new boolean resource allowCustomHosts


src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -213,40 +225,85 @@ class ServerMetadata {
...
+						result.put(url, name);
+					}
+				}
+				eventType = xmlParser.next();
+			}
+
+			return result;
+		} catch (XmlPullParserException | IOException e) {
+			e.printStackTrace();
+			return new HashMap<>();
```
jkuester, line 302:
```suggestion
			return Collections.emptyMap();
```

Super minor, but avoids another object on the heap.... (And just as a general best practice I like to use this when I know the map will always be empty.)

src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -213,40 +225,85 @@ class ServerMetadata {
...
+
 		SharedPreferences.Editor ed = prefs.edit();
-		ed.putString(url, friendly(url));
+		ed.putString(url, name);
 		ed.apply();
 	}
 
+	private static Map<String, String> parseInstanceXML(Context context) {
+		try {
+			HashMap<String, String> result = new HashMap<>();
```
jkuester, line 281: Isn't it great that Java does not have tuples!   :)  Android utils does have a `Pair` class, but FWIW I am not really sure it would improve much to make this `List<Pair<String, String>> result...`

src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
```diff
@@ -213,40 +225,85 @@ class ServerMetadata {
...
+				if (eventType == XmlPullParser.START_TAG) {
+					String tagName = xmlParser.getName();
+					if (tagName.equals("instance")) {
+						String name = xmlParser.getAttributeValue(null, "name");
+						String url = xmlParser.nextText();
+						result.put(url, name);
+					}
+				}
+				eventType = xmlParser.next();
+			}
```
jkuester, line 297:
```suggestion
			while (xmlParser.next() != XmlPullParser.END_TAG) {
				if (xmlParser.getEventType() != XmlPullParser.START_TAG
					|| !"instance".equals(xmlParser.getName())) {
					continue;
				}
				String name = xmlParser.getAttributeValue(null, "name");
				String url = xmlParser.nextText();
				result.put(url, name);
			}
```
It pains me that the `XmlResourceParser` is the recommended (only?) way to parse xml data from `R.xml`.  :cry: Such a primitive API!!!

I am not going to waste any more time trying to fine an alternative without a while-loop, but here is a slightly less nested version that I think should still work!

jkuester (changes requested):
Sorry for the delayed response here!  All in all a great approach!  Just added some minor thoughts inline.

jkuester:
@ngaruko I know you were involved with a bunch of the early testing for this (to build the initial APK).  However, this work has mostly aligned with the Allies (and we have been reviewing the code).  Would you like to give the final QA blessing on this PR or should I bring @tatilepizs up to speed here?  

I am curious to get QA's perspective on the cht-android process! At present we don't have much in terms of automated e2e tests or anything.  Should that be a priority for this PR?

src/main/java/org/medicmobile/webapp/mobile/SettingsDialogActivity.java
kennsippell: ya surprising. i checked out JSON also and it looks like it sucks an equal amount

kennsippell:
@jkuester Thanks for the constructive and excellent feedback. Much appreciated.

jkuester (approved):
LGTM!! :sunglasses:

src/moh_kenya_echis/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+	<string name="app_name">eCHIS Kenya</string>
+	<bool name="allowCustomHosts">False</bool>
```
m5r, line 4: should we add this property to every project? or if the fallback value when building a branded app is always `False`, should we remove it here?

m5r (approved):
This is looking good; I just left one question but not worth blocking the merge once answered

src/moh_kenya_echis/res/values/strings.xml
jkuester: So, `allowCustomHosts` only is relevant if you _do not_ have an `app_host` value.  For normal brands that only target a single CHT server, you should put the URL for the servers as the `app_host` value. When `app_host` exists, then the server-list page (for choosing which server to connect to) is never shown.  

To be able to use your branded app (or the Unbranded app) to connect to _multiple different CHT servers_, you should not set the `app_host` value.  That will cause the Server-list page to be displayed when the app is run for the first time.  The `allowCustomHosts` flag just controls whether or not the server-list will allow users to enter a custom URL (instead of just picking one from this list).

We want the _Unbranded_ app to allow custom hosts. So, in the `main` config we set `allowCustomHosts = true` (since that is where the unbranded app gets all its config).  Any other brand that has an `app_host` value (so all of them besides `moh_kenya_echis`) does not need to set a value.

m5r: It makes a lot of sense, thanks a lot for the detailed explanation @jkuester!

PR #320, opened 2023-06-09
fix(#319): launch proper server from filtered server list

### LINKED ISSUE #319: Server list entries do not launch proper entry after filtering with search box
**Describe the bug**
When using the server list, the search box will filter the entries in the list.  However, the targets that get launched when clicking on the button do not change based on the filtering!  

**To Reproduce**
1. Open the Unbranded app
2. See that "Custom" is the first entry
3. Enter "Gamma" into the search box
4. See that now "Gamma Dev" is the first entry in the list
5. Click "Gamma Dev"
6. See that instead of launching the Gamma Dev instance, you end up in the Custom URL screen

**Expected behavior**
The targets for the buttons should match the displayed button!

**Environment**
- App Version: `master`

jkuester:
Closes #319 

## Testing considerations

These changes affect the logic that determines which servers are launched when an item is selected in the server list.  We need to check that the expected server is launched.  The custom URL logic should also be tested to ensure the custom URL page is launched only when expected.  

## Apks
[unbranded.zip](https://github.com/medic/cht-android/files/11709631/unbranded.zip)
[Uploading moh_kenya_echis.zip…]()

### DIFF
--- a/src/androidTestUnbrandedDebug/java/org/medicmobile/webapp/mobile/SettingsDialogActivityTest.java
+++ b/src/androidTestUnbrandedDebug/java/org/medicmobile/webapp/mobile/SettingsDialogActivityTest.java
@@ -160,7 +160,7 @@ public class SettingsDialogActivityTest {
 	}
 
 	@Test
-	public void testFilterServers() {
+	public void testFilterServers() throws InterruptedException {
 		onView(withId(R.id.lstServers)).check(matches(isDisplayed()));
 		onView(withText("Custom")).check(matches(isDisplayed()));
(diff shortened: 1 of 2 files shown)


jkuester:
cc @kennsippell @derickl @mrjones-plip

tatilepizs (approved):
It works perfectly! 🎉 

[video](https://github.com/medic/cht-android/assets/94494491/605dc520-4f92-48d4-b563-20e6e654eea8)

latin-panda (approved):
LGTM!

Issue #324, opened 2023-08-20
Update target API to latest


garethbowen:
From Google Play Store

> App must target Android 13 (API level 33) or higher
> To provide users with a safe and secure experience, Google Play requires all apps to meet target API level requirements.
> From 31 Aug 2023, if your target API level is not within one year of the latest Android release, you won't be able to update your app.

In other words, after August 31 we won't be able to publish updated (or presumably new) android apps until the target API is at least 33.

This has the potential to block new deployments going live so we need to fix this as soon as possible before then.

ngaruko:
Testing apk from [PR](https://github.com/medic/cht-android/pull/325) :
<details>
 <summary> App installed successfully</summary>
 | Android 10  (SDK 29 ) | Android 12  (SDK 31) | Android 13  (SDK 33) |
| ------------- | ------------- | ------------- |
|  
![image](https://github.com/medic/cht-android/assets/6979995/07f2d72f-6ea0-4bd5-b9fb-a315b6fc9b49)
 |  
<img width="268" alt="image" src="https://github.com/medic/cht-android/assets/6979995/bda604b5-b483-47de-a831-2f165df65e31">
 | 
![image](https://github.com/medic/cht-android/assets/6979995/56bcc4ac-a6d8-4d42-9ef9-04d2ad15e54d)
 |
</details>

latin-panda:
@ngaruko I found that to test sending reports by SMS, the permission needs to be enabled in the [manifest](https://github.com/medic/cht-android/blob/2061adccdb52b912a0ed79361284713e8494d71a/src/main/AndroidManifest.xml#L26), by default is commented. 

<details>
<summary>Screenshot of SMS that was sent after submitting report</summary>

<img src="https://github.com/medic/cht-android/assets/66472237/695e2ce0-0016-41f6-86e6-3a9f3cca4836" width="300">
</details>

latin-panda:
@ngaruko I did these same [cases you tested](https://github.com/medic/cht-android/issues/324#issuecomment-1692724352) in Android 9 and 13, with these additional cases:
- Back button, no closing the app
- Training cards
- [Reports sending SMS](https://github.com/medic/cht-android/issues/324#issuecomment-1692941001) (on Android 13 only)

ngaruko:
@latin-panda I have noticed the buck button not closing the app. Did the app launcher with the old MRDT app on Android 13 and 12. All good.

garethbowen:
@latin-panda Can we close this now?

Issue #159, opened 2021-02-02
Replace deprecated ImagePicker


garethbowen:
The ImagePicker library we use is [deprecated](https://github.com/Mariovc/ImagePicker#imagepicker-deprecated-use-google-api-camerax), and the most recent version [crashes the app](https://github.com/medic/medic-android/issues/127), so we should migrate to another solution.

garethbowen:
Scheduling for 4.0.0 because some of the solutions will not work on Android 4.4.

latin-panda:
Adding a bit more of context:

Currently the users can't upload image, audio, video and file when using the cht-android app. This work will fix these type of questions and increase the compatibility with the [ODK standard](https://xlsform.org/en/ref-table/).

latin-panda:
To add more info: 
Webapp's Enketo has a max limit of 32KB for each uploaded file. There's another ticket to increase this limit [here](https://github.com/medic/cht-core/issues/7529).

latin-panda:
Ready for AT in [this PR](https://github.com/medic/cht-android/pull/245)

Please see the current limitations in[ the previous comment,](https://github.com/medic/cht-android/issues/159#issuecomment-1046735496) and send me any question you might have 🙂

This is one form we've been using to test, please include more cases as needed for testing.
<details>
<summary>xml form</summary>


```

<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml">
	<h:head>
		<h:title>Photos on Android - Demo Form</h:title>

		<model>
			<instance>
				<photos-on-android id="photos-on-android">
					<my_photo/>
					<my_video/>
					<my_sound/>
					<my-any-file/>
					<meta>
						<instanceID/>
					</meta>
				</photos-on-android>
			</instance>

			<bind nodeset="/photos-on-android/my_photo" type="binary"/>
			<bind nodeset="/photos-on-android/my_video" type="binary"/>
			<bind nodeset="/photos-on-android/my_sound" type="binary"/>
			<bind nodeset="/photos-on-android/my-any-file" type="binary"/>
		</model>
	</h:head>

	<h:body>
		<upload ref="/photos-on-android/my_photo" mediatype="image/*">
			<hint>Select a picture or take a photo</hint>
			<label>Image widget</label>
		</upload>
		<upload ref="/photos-on-android/my_video" mediatype="video/*">
			<hint>Select a video</hint>
			<label>Video widget</label>
		</upload>
[...]

Issue #322, opened 2023-07-19
Unresolvable SMS permission prompt when submitting reports for CHT < 3.9


jkuester:
## To recreate the behavior

- On Android 6+ device, open cht-android `1.0.0+` app using a brand that does not have the [`SEND_SMS` permission enabled](https://github.com/medic/cht-android/blob/b2e815fe383655156871a702980a4ca7a731ef8a/src/main/AndroidManifest.xml#L26)
    - Unbranded version does not have `SEND_SMS` permission enabled
- Connect to a CHT instance that is `< 3.9.0`
- Submit app form that does not have any [`xml2sms` configuration](https://docs.communityhealthtoolkit.org/apps/guides/forms/app-form-sms/#sending-reports-as-sms) set
- See prominent disclosure page for SMS permission
- Press "Allow"
- No prompt for SMS permission is displayed. Instead the Android system settings for the CHT app are opened. 
    - You have to press the back button to return to the CHT app

Basically, the [sending report as SMS](https://docs.communityhealthtoolkit.org/apps/guides/forms/app-form-sms/#sending-reports-as-sms) functionality is getting triggered unintentionally and this is inadvertently triggering the SMS permission request which fails because the app does not have `SEND_SMS` permission enabled.

## Details

Due to https://github.com/medic/cht-core/issues/6162, submitting a report in cht-android for CHT versions `< 3.9.0` will always trigger an SMS record for the report to be sent regardless of the `xml2sms` configuration for the form.  In older versions of cht-android, if your brand was not configured with the `SEND_SMS` this send-attempt would always cause an error that was handled silently by the app (but visible in the adb logs). 

In cht-android `1.0.0`, we [updated handling](https://github.com/medic/cht-android/pull/262) for the `SEND_SMS` permission to properly request permission from the user when trying to send a report via sms. BUT this only works when the `SEND_SMS` permission has been [configured for the app brand](https://github.com/medic/cht-android/blob/b2e815fe383655156871a702980a4ca7a731ef8a/src/main/AndroidManifest.xml#L26).  The idea is that to send reports via sms this is what is needed:

1. Publish cht-android app with `SEND_SMS` permission registered
2. [Configure form](https://docs.communityhealthtoolkit.org/apps/guides/forms/app-form-sms/#sending-reports-as-sms) with `xml2sms` value
3. User approves permission request when prompted by app (should happen the first time the app tries to send a report via sms). 

[...]

PR #325, opened 2023-08-23
chore(#324): upgrading to android 13

latin-panda:
**Description**

This is to upgrade from Android 11 to Android 13.
- READ_EXTERNAL_STORAGE is ignored. Therefore, we stopped requesting this permission for Android +13 with the prominent disclosure prompt. At the same time, we will keep asking this permission for any device in previous versions of Android (12, 11, 10, etc). 
  - All devices regardless of the Android version should be able to get files (png, video, audio, pdf, xls etc.) using [File Picker](https://github.com/medic/cht-android/issues/159) and [Android App Launcher](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#android-app-launcher)
- We limited the SDK target on the following tests because our current test dependencies don't support target 33
  - SmsSenderTest.java 
  - FilterableListAdapterTest.java
  - AsyncExecutorTest.java
  - VibratorTest.java

#324

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/RequestStoragePermissionActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/RequestStoragePermissionActivity.java
@@ -30,13 +30,6 @@ public class RequestStoragePermissionActivity extends FragmentActivity {
 		registerForActivityResult(new ActivityResultContracts.RequestPermission(), isGranted -> {
 			Intent responseIntent = createResponseIntent();
 
-			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-				trace(this, "RequestStoragePermissionActivity :: READ_EXTERNAL_STORAGE permission is ignored in Android 13+.");
-				setResult(RESULT_OK, responseIntent);
-				finish();
-				return;
-			}
-
 			if (isGranted) {
 				trace(this, "RequestStoragePermissionActivity :: User allowed storage permission.");
 				setResult(RESULT_OK, responseIntent);
@@ -80,6 +73,13 @@ public class RequestStoragePermissionActivity extends FragmentActivity {
 	@Override
 	public void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
+
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+			trace(this, "RequestStoragePermissionActivity :: READ_EXTERNAL_STORAGE permission is ignored in Android 13+.");
+			setResult(RESULT_OK, createResponseIntent());
+			finish();
+			return;
+		}
 
 		this.requestWindowFeature(Window.FEATURE_NO_TITLE);
 		setContentView(R.layout.request_storage_permission);
--- a/src/main/java/org/medicmobile/webapp/mobile/RequestStoragePermissionActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/RequestStoragePermissionActivity.java
@@ -30,6 +30,13 @@ public class RequestStoragePermissionActivity extends FragmentActivity {
 		registerForActivityResult(new ActivityResultContracts.RequestPermission(), isGranted -> {
 			Intent responseIntent = createResponseIntent();
 
+			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+				trace(this, "RequestStoragePermissionActivity :: READ_EXTERNAL_STORAGE permission is ignored in Android 13+.");
+				setResult(RESULT_OK, responseIntent);
+				finish();
+				return;
+			}
+
 			if (isGranted) {
 				trace(this, "RequestStoragePermissionActivity :: User allowed storage permission.");
 				setResult(RESULT_OK, responseIntent);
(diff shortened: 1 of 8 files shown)


src/main/java/org/medicmobile/webapp/mobile/RequestStoragePermissionActivity.java
```diff
@@ -74,6 +74,13 @@ public class RequestStoragePermissionActivity extends FragmentActivity {
 	public void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+			trace(this, "RequestStoragePermissionActivity :: READ_EXTERNAL_STORAGE permission is ignored in Android 13+.");
+			setResult(RESULT_OK, createResponseIntent());
+			finish();
```
latin-panda, line 80: Just in case if someone is wondering

> You can call finish() from within this function, in which case onDestroy() will be immediately called after onCreate(Bundle)

[source](https://developer.android.com/reference/android/app/Activity#onCreate(android.os.Bundle))

src/test/java/org/medicmobile/webapp/mobile/SmsSenderTest.java
```diff
@@ -46,6 +46,7 @@
 import java.util.stream.IntStream;
 
 @RunWith(RobolectricTestRunner.class)
+@Config(sdk = 31) // ToDo: Remove when upgrading robolectric
```
latin-panda, line 49: We're getting this error because our current version of `robolectric` doesn't support `target 33`

```
java.lang.IllegalArgumentException at RobolectricTestRunner.java:260
        Caused by: java.lang.IllegalArgumentException at DefaultSdkPicker.java:118
```
In this [link](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/docs/testing.md#robolectric), they recommend to use `Config(sdk = 31)` to make the test pass as a _temporary fix_.

latin-panda:
@garethbowen, it turns out that our version of `robolectric` doesn't fully support target 33. 
I tried upgrading the library but it has deprecations that we need to fix and it requires upgrading `androidx` libraries, which also have more deprecations that we need to figure out. It's going to take longer to resolve them. 

Knowing that we have limited time before PlayStore blocks our apps, I've dropped the work on the tests and patch them by running them in target 31. I created a separate ticket to [fix the dependencies](https://github.com/medic/cht-android/issues/326).  Then @ngaruko and I can focus on testing.

My question is:

If the manual testing is good, can we release this to unblock the PlayStore and work on the dependencies later?

garethbowen (approved):

garethbowen:
@latin-panda Yes let's go manual to unblock this - incomplete testing is better than blocking publishing altogether. But let's make sure we fix this soon - if you don't have time to take this on let me know and I'll shop it around.

Thanks for doing such a deep dive here!

ngaruko (commented):
1. Test most critical [scenarios](https://github.com/medic/cht-android/issues/324#issuecomment-1692724352) and they all passed
2. Random testing of a few workflows
3. Reading from Google developer, one of the major changes around permissions is replacing `[READ_EXTERNAL_STORAGE](https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE)` by [granular media permission](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions). Apparently we removed the permission and don't seem to have introduced any granular permission. Is that intentional @latin-panda ?

latin-panda:
> Reading from Google developer, one of the major changes around permissions is replacing [READ_EXTERNAL_STORAGE](https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE) by [granular media permission](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions). Apparently we removed the permission and don't seem to have introduced any granular permission. Is that intentional @latin-panda ?

@ngaruko I saw that, but I'm not clear under what circumstances it's needed; for example, they don't mention anything about the other types of files that aren't images, video, or audio, such as pdf or spreadsheets, and that we were accessing using `READ_EXTERNAL_STORAGE` (sample [form](https://github.com/medic/support-scripts/tree/master/forms/enketo_widgets)). 
Some blogs say that since it's in a shared place then it's not needed. When I tested it, it worked on my phone even when not requesting any of those permissions. 
- Are you experiencing any issue using the using https://github.com/medic/cht-android/issues/159 and [Android App Launcher](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#android-app-launcher) when accessing any type of file?
- How else can we test this and get feedback? Maybe we're missing they way to reproduce it.

In another topic: 
- Did you try the feature of sending SMS when submitting forms? I need help getting the form configured for that. There's a known [issue](https://github.com/medic/cht-android/issues/322), but we just don't want to make it worse

garethbowen:
@latin-panda I don't have a lot of android experience either but I think your "reason 3" is spot on. Thanks for being so diligent!

latin-panda:
@ngaruko waiting for your approval or feedback after you've done verifying the app

ngaruko:
@latin-panda . Thanks for all the research. It all makes sense and practically, I do not see any scenario where access is denied due to the permission not present. So, good to go.

ngaruko (approved):
Good research on those permissions. 👍

PR #331, opened 2023-10-04
Add digital payments flavour

eljhkrr:


jkuester (approved):
This looks good!  The only question I have is about the brand name.  

I don't have a lot of context for this particular app, but I see from the strings.xml that is it for Zanzibar.  In that case, it feels like `digital_payments` is a bit vague.  Unless we are planning to have this app reused by other deployments, I would have expected a brand name more like `moh_zanzibar_payments` or something....  (But, like I said, I don't have a lot of context here, so if the branding has already been discussed, then I do not intend to re-open the conversation here!)

Also, this is minor, but do we have any icon to supply for this APK?

eljhkrr:
Thanks for the feedback. We don't have any icons yet for the project and the naming was based off the main instance name for the pilot. Once it goes live the preferred name would be `moh_zanzibar_payments`.

PR #333, opened 2023-10-19
adds vhtapp_uganda bundle

jonathanbataire:


jonathanbataire:
> The problem here is again with the keys... we should not have access to the signing key so that means we can't build and sign the app for MoH. The approach that would give full ownership would be for MoH to set up their own build pipeline/process to publish the app themselves.

okay cool let me work with them on that

PR #334, opened 2023-10-23
feat(#334): add apk for cares malawi

freefony:


jkuester (approved):
All the code changes look great!

Regarding the test failure, I am completely baffled.  I believe it is failing [here](https://github.com/medic/cht-android/blob/d1c65166aa7050605aac0aaeb41f4fdd19782c73/src/androidTestUnbrandedDebug/java/org/medicmobile/webapp/mobile/SettingsDialogActivityTest.java#L109) when it is waiting for the Gamma login screen to show up.  

> 'has element with id: form' doesn't match: <?xml version="1.0" encoding="UTF-8"?><html xmlns:html="http://www.w3.org/1999/xhtml">

You can see from the error message that the actual page displayed is just the initial loading page for the webapp (with the loading spinner), but for some reason the loading never finishes and the login page is not displayed. Currently the test is waiting `7sec` for the login screen to load.  I tried [increasing that to `2min`](https://github.com/medic/cht-android/pull/335) but the test still fails.  

I have not been able to reproduce this issue locally, however. The test runs fine for me, both on its own when I run it in Android Studio and when I run it via `make test-ui`.

I can reproduce the issue [on top of `master`](https://github.com/medic/cht-android/pull/336) though.  So, it is clearly not related to this PR at all...  

If nobody has any brilliant ideas on what could be going wrong with this test, we may just have to log a new issue to get it fixed and skip it in the code for now.  There is not much sense in it being a huge blocker for this PR, but I am fresh out of ideas for what could be causing the problem....

garethbowen:
I agree with @jkuester - the failure probably has nothing to do with this PR. Let me know when it's ready and I can force merge this to get past the blocking required build.

@jkuester Would you mind raising an issue about the test failure? So far what I have is...

- Slightly sus looking `[DDMLIB]: An unexpected packet was received before the handshake.` errors. But these are probably not causing this failure.
- It's trying to log in to gamma-cht.dev (I think) which is in a bad state. Firstly the admin password in 1pass doesn't work so I can't log in. Secondly the /api/v1/monitoring URL 404s which implies it's a really old instance. It should be wiped and reset to 4.4 or deleted from the android app picker. A quick test would be to change the test to try logging in to "gamma.dev" instead (list item 2).
- As Josh pointed out it's loading the spinner page. To clarify, it's the spinner page from the webapp _not_ the login page. What should happen is that page loads, the JS downloads, and then the JS executes, which detects there's no session, and redirects to the login page. Any one of these steps could be failing. It's possible the instance is just bricked, or that we need to [enable javascript](https://developer.android.com/training/testing/espresso/web#javascript-support), or we need to allow redirects, or we just need to give it more time.

jkuester:
https://github.com/medic/cht-android/issues/337

I do not see how it could be a Gamma problem since it works fine locally.  (And even after a 2min wait on the CI, it never made it to the login page...)

PR #341, opened 2023-12-08
fix: upgrading ruby version for fastlane

latin-panda:


binokaryg:
I am [getting errors](https://github.com/medic/cht-android/actions/runs/9692429886/job/26745745380#step:6:21) with Ruby 2.7 also. It wants `>= 3` now.

> ERROR:  Error installing fastlane:
	The last version of public_suffix (>= 2.0.2, < 7.0) to support your Ruby & RubyGems was 5.1.1. Try installing it with `gem install public_suffix -v 5.1.1` and then running the current command again
	public_suffix requires Ruby version >= 3.0. The current ruby version is 2.7.8.225.

PR #342, opened 2023-12-12
Updated keystores for CARES APKs

kennsippell:
Seems the keystores generated in https://github.com/medic/cht-android/pull/334 do not match the secrets in 1password. This regenerates and commits fresh keystores.


freefony (approved):
Looks like deleted files only.

Issue #337, opened 2023-11-09
Login Test Broken


jkuester:
**Describe the issue**
Our [e2e test](https://github.com/medic/cht-android/blob/d1c65166aa7050605aac0aaeb41f4fdd19782c73/src/androidTestUnbrandedDebug/java/org/medicmobile/webapp/mobile/SettingsDialogActivityTest.java#L109) for loading the Gamma login page is broken.  The tests run fine for me locally, but they fail on CI.

The spinner page from the webapp is loaded into the webview. What should happen is that page loads, the JS downloads, and then the JS executes, which detects there's no session, and redirects to the login page. However, when running on CI, it never makes it to the login page. Any one of these steps could be failing.

More details/discussion on: https://github.com/medic/cht-android/pull/334

**Describe the improvement you'd like**
The test shalt pass.

Issue #345, opened 2024-01-16
CHT Android repository needs to find way around 100 secret/repo limitation


mrjones-plip:
**Describe the issue**
MoH Togo can not publish their APK because this cht-android repo has hit the [100 secret/repo limit](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#limits-for-secrets). They're looking for an immediate term work around for this:

* https://github.com/medic/cht-android/issues/344

**Describe the improvement you'd like**
A long term solution so that future deployments have an easy way to publish their APKs. This might include:
* [Using AWS Secrets in GH Actions](https://docs.aws.amazon.com/secretsmanager/latest/userguide/retrieving-secrets_github.html)
* Using a single encrypted file with 100s of secrets in it. Something [GPG might work](https://blog.mikepenz.dev/a/manage-secrets-ga/). We'd need to be sure we weren't logging secrets though, so [proper handling/masking should be used](https://github.com/actions/toolkit/tree/main/packages/core#setting-a-secret) in this case to avoid it being logged in cleartext.

**Describe alternatives you've considered**
* Have each new deployment beg other APKs to rescind their secret's from this repository, as [Togo is doing now](https://github.com/medic/cht-android/issues/344)

kennsippell:
There are 24 secrets for `_ALIAS_` and `_KEYSTORE_PATH_` information which I believe are totally unused and might provide move  months/years of needed secrets.

mrjones-plip:
@craig-landry  or @garethbowen - I can't see the secrets (not the values obviously, but the number of them and their names).  I think you're both busy right now (this week?), but when you're "back in the office", can you look into a basic audit of the secrets we have in the repo? 

Maybe getting rid of the 24 Kenn mentioned is enough to wrap this ticket up?

garethbowen:
Another option is to use the 1pass API to store them there. We could have a vault dedicated to this.

> can you look into a basic audit of the secrets we have in the repo?

I did a quick search on the code and I think those are referenced, eg: https://github.com/medic/cht-android/blob/5e64289fe74a125f0118cdd7daaaa04833ae1ccc/Makefile#L173-L174

However why this would ever be set to anything other than the default I don't know.

@kennsippell Any ideas? I'm guessing these secrets all have the same value so it'd be safe to delete?

kennsippell:
[...]

PR #347, opened 2024-02-19
fix(#345): Hardcode the values `ANDROID_KEY_ALIAS` and `ANDROID_KEYSTORE_PATH` in make command `check-env`

sugat009:
The changes include hardcoding the values `ANDROID_KEY_ALIAS` AND `ANDROID_KEYSTORE_PATH` in the `make` command `check-env` such that they don't use the secrets in the repository.

Fixes: (#345 )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,10 +168,8 @@ ifndef ANDROID_SECRETS_IV
 	$(eval ANDROID_KEYSTORE_PASSWORD := $(shell echo ${${VARNAME}}))
 	$(eval VARNAME=ANDROID_KEY_PASSWORD_${ORG_UPPER})
 	$(eval ANDROID_KEY_PASSWORD := $(shell echo ${${VARNAME}}))
-	$(eval VARNAME=ANDROID_KEY_ALIAS_${ORG_UPPER})
-	$(eval ANDROID_KEY_ALIAS := $(shell echo ${${VARNAME}}))
-	$(eval VARNAME=ANDROID_KEYSTORE_PATH_${ORG_UPPER})
-	$(eval ANDROID_KEYSTORE_PATH := $(shell echo ${${VARNAME}}))
+	$(eval ANDROID_KEY_ALIAS := "medicmobile")
+	$(eval ANDROID_KEYSTORE_PATH := "${org}.keystore")
 endif
 
 check-keystore-exist:
@@ -216,8 +214,6 @@ secrets/secrets-${org}.tar.gz.enc: secrets/secrets-${org}.tar.gz
 	$(info export ANDROID_KEY_PASSWORD_$(ORG_UPPER)=$(ANDROID_KEYSTORE_PASSWORD))
 	$(info export ANDROID_SECRETS_IV_$(ORG_UPPER)=$(ANDROID_SECRETS_IV))
 	$(info export ANDROID_SECRETS_KEY_$(ORG_UPPER)=$(ANDROID_SECRETS_KEY))
-	$(info export ANDROID_KEYSTORE_PATH_$(ORG_UPPER)=$(ANDROID_KEYSTORE_PATH))
-	$(info export ANDROID_KEY_ALIAS_$(ORG_UPPER)=$(ANDROID_KEY_ALIAS))
 	$(info )
 	$(info #)
 	$(info # The file secrets/secrets-${org}.tar.gz.enc was created and has to be added to the git)
--- a/src/test/bash/test-keystore.bats
+++ b/src/test/bash/test-keystore.bats
@@ -88,8 +88,6 @@ teardown() {
   export ANDROID_KEY_PASSWORD_TEST="abc"
   export ANDROID_SECRETS_IV_TEST="1234abc"
   export ANDROID_SECRETS_KEY_TEST="111222"
-  export ANDROID_KEYSTORE_PATH_TEST="test.keystore"
-  export ANDROID_KEY_ALIAS_TEST="medicmobile"
   # Now trying to decrypt without the right env sets fails
   run make org=test keydec
   assert_failure 2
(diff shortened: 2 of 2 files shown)


kennsippell:
Nice work @sugat009. I suspect this is the simplest functional version of the change possible.

I recommend you now

1. Get tests passing
2. Test that publication still works for all flavors.  For this, I'd recomment you publish an "alpha release". I can't find the documentation explaining how to do that, so recommend you ask on forum or #development.

sugat009:
The instrumentation tests seem to be failing for this PR but I'm not sure if these changes have anything to do with the test failing. After going through the already existing issues and PR, the issue seems to be persisting even in the `main` branch and an issue has been created to address it as well #337 . So, I guess we can move on with the action items for this PR.
Ref: https://github.com/medic/cht-android/pull/334#pullrequestreview-1707062417
Ref: https://github.com/medic/cht-android/pull/334#issuecomment-1802634663

Makefile
```diff
@@ -168,10 +168,8 @@ ifndef ANDROID_SECRETS_IV
 	$(eval ANDROID_KEYSTORE_PASSWORD := $(shell echo ${${VARNAME}}))
 	$(eval VARNAME=ANDROID_KEY_PASSWORD_${ORG_UPPER})
 	$(eval ANDROID_KEY_PASSWORD := $(shell echo ${${VARNAME}}))
-	$(eval VARNAME=ANDROID_KEY_ALIAS_${ORG_UPPER})
-	$(eval ANDROID_KEY_ALIAS := $(shell echo ${${VARNAME}}))
-	$(eval VARNAME=ANDROID_KEYSTORE_PATH_${ORG_UPPER})
-	$(eval ANDROID_KEYSTORE_PATH := $(shell echo ${${VARNAME}}))
+	$(eval ANDROID_KEY_ALIAS := "medicmobile")
+	$(eval ANDROID_KEYSTORE_PATH := "${org}.keystore")
```
jkuester, line 172: Can remove these doc lines down below:

```
	$(info export ANDROID_KEYSTORE_PATH_$(ORG_UPPER)=$(ANDROID_KEYSTORE_PATH))
	$(info export ANDROID_KEY_ALIAS_$(ORG_UPPER)=$(ANDROID_KEY_ALIAS))
```

jkuester: Also, should be able to remove these variables from the [test-keystore.bats](https://github.com/medic/cht-android/blob/5e64289fe74a125f0118cdd7daaaa04833ae1ccc/src/test/bash/test-keystore.bats#L91) file (make sure to remove from both tests) without those tests failing now!

jkuester (commented):
:+1: I think Kenn is spot on here!  These changes look good to me (have just requested a few additional clean-ups below).  Before merging this PR, lets follow Kenn's suggestion and try to run a tag build (basically just going to be steps `#2`-`3` of the [Alpha for release testing steps](https://docs.communityhealthtoolkit.org/contribute/code/android/releasing/#alpha-for-release-testing) except we are just going to do it in the branch before even going to master).  Should be able to create a tag right on the HEAD of this branch called `v1.1.2-secrets-345` and push it to GitHub.  Then watch to see if the [publish](https://github.com/medic/cht-android/actions/workflows/publish.yml) workflow succeeds.  If it does, then I think we are good to go and can merge this PR.

> 5. Commit and tag as latest stable

@kennsippell Is there a reason we need a new tag at this time?  (These changes seem to just be with the build process and so will be needed for the next release, but I guess we don't need to actually release anything right now...)

Also, regarding the test failure, since we have not fixed https://github.com/medic/cht-android/issues/337, I have no reason to expect that test to pass.  :disappointed: Since nothing is actively blocked yet by this PR, I think we need to bite the bullet and try to solve #337 for this PR :grimacing:.  The sad fact is that if we don't do it now, it may never get done. :sweat: 

The bad news for that is that I pretty much exhausted all my ideas for that test with the investigation/documentation that I already did.  @sugat009 how much experience do you have with Android tests? :sweat_smile:   Maybe you can see what I missed.  Otherwise, my next best idea is to maybe just try and upgrade our way out of the problem (always a winning strategy).  Some ideas to try (from most to least promising):

- build.yml:
    - I notice that our targetSdk in the build.gradle is `33`, but the api-level in our CI build.yml jobs for running the tests is `29`.  Maybe bumping that will help.  
    - Can also try bumping the image versions used by those CI job (if we are not on the latest)
- build.gradle
    - Updating the versions of our `androidTestImplementation` dependencies (where possible)

kennsippell:
> Is there a reason we need a new tag at this time? (These changes seem to just be with the build process and so will be needed for the next release, but I guess we don't need to actually release anything right now...)

Don't need to release per se, but all future releases will need to be based off this. Once the secrets are gone, old versions will not be publishable.  I just noticed this comment 

> Check out the tag from the [last stable release](https://github.com/medic/cht-android/releases) in CHT Android repository and create a branch
https://docs.communityhealthtoolkit.org/apps/guides/android/branding/#2-new-brand

So my comment about the stable tag is just to ensure it is the base branch for future work.

sugat009:
@jkuester I've made the changes requested and also did the steps from the `alpha-release` by creating a tag(`v1.1.2-secrets-test`) (`v1.1.2-secrets-345` tag makes the CI fail due to a constraint in the `build.gradle` file). The [pipeline](https://github.com/medic/cht-android/actions/runs/7986413489) is passing now for the tag-push atleast. 😅 
@kennsippell Once the PR is approved the secrets can hopefully be deleted.

@jkuester @lorerod Please review.

> The bad news for that is that I pretty much exhausted all my ideas for that test with the investigation/documentation that I already did. @sugat009 how much experience do you have with Android tests? 😅 Maybe you can see what I missed. Otherwise, my next best idea is to maybe just try and upgrade our way out of the problem (always a winning strategy). Some ideas to try (from most to least promising):

@jkuester Unfortunately, I have no experience in java or android at all. 😅  I guess this needs to be done another time separately as this issue will be prolonged.

sugat009:
@lorerod In my opinion, it would be preferable to utilize the existing issue #337 to tackle the failing tests in a different PR. Addressing this concern within the current task would lead us to deviate from its intended scope.

jkuester (approved):
LGTM!  

In other good news, I think I got the CI issues fixed over in https://github.com/medic/cht-android/pull/348.

jkuester:
@kennsippell when you get the chance, can you remove the unneeded secrets from the repo now?

kennsippell:
@jkuester @sugat009 I've deleted unneeded secrets. There are now only 75 tokens in the cht-android repository.

PR #348, opened 2024-02-23
chore(#337): fix e2e tests on CI

jkuester:
Closes https://github.com/medic/cht-android/issues/337

Upgrade our way out of the CI test failures. Uplifts all our `androidTestImplementation` dependencies to the latest versions. Also updates our GitHub actions jobs for running the tests to use Ubuntu and build against Android 33.  This has the added benefit of making the tests run quite a bit faster for some reason...


latin-panda (approved):
Thanks Josh! It's good to go

.github/workflows/build.yml
```diff
@@ -75,7 +75,7 @@ jobs:
   instrumentation-tests:
 
     name: Instrumentation tests
-    runs-on: macos-12
+    runs-on: ubuntu-22.04
```
m5r, line 78: I remember having a conversation with Mariano about why we're using MacOS to run our android e2e tests. Back then, the android emulator had [terrible support for linux](https://github.com/ReactiveCircus/android-emulator-runner/issues/81#issuecomment-658291981)
Is it good enough to switch to linux now?

jkuester: Ah, thanks for the context @m5r!  I had not been able to understand why this was originally on macOS.  The good news is that since the GH Linux runners [now support KVM hardware acceleration](https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/), the Linux runners are actually [much faster for Android builds](https://github.com/ReactiveCircus/android-emulator-runner/issues/46#issuecomment-1899848624) than macOS (this improvement was noticeable even for our tests).

jkuester: Also worth noting that the [example usage documentation](https://github.com/ReactiveCircus/android-emulator-runner?tab=readme-ov-file#usage--examples) for `ReactiveCircus/android-emulator-runner` now suggest using Linux.

PR #352, opened 2024-03-22
chore: Add sonar configuration

jkuester:
Adding sonar configuration as part of the [process for enabling Sonar](https://docs.communityhealthtoolkit.org/contribute/code/static-analysis/#adding-a-new-repo-to-sonarcloud) on a new repo.

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,11 @@
+sonar.sources=.
+# Can have multiple comma-separated entries
+sonar.exclusions=**/test*/**/*, **/androidTest*/**/*
+
+sonar.tests=.
+# Can have multiple comma-separated entries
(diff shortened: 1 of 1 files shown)


mrjones-plip (approved):
:lab_coat: !!

spot checked docs and compared to CHT Core's version - looks good!

not at all blocking merge, but do we need to [specify java version](https://docs.sonarsource.com/sonarqube/latest/analyzing-source-code/languages/java/#java-source-version) like we specify [python version](https://github.com/medic/cht-core/blob/ed4b081453d70b064430e2d82cc0bd3ff3b5b13f/.sonarcloud.properties#L10)?

jkuester:
I do not _think_ we need to specify the Java version since Sonar should be about to load it from the `build.gradle` config.  At least I hope so, since I do not really want another place to have to update whenever we bump the Java version...

PR #353, opened 2024-03-27
fix(#308): Add autoverify=true to AndroidManifest.xml file

### LINKED ISSUE #308: Deep links broken by default on Android 12+
**Describe the bug**
Android 12 raised the bar on what is required for automatically providing deep links to your application (via intents).  TLDR is that [getting it to work](https://doordash.engineering/2022/01/25/your-deep-links-might-be-broken-web-intents-and-android-12/) will require deploying a file with a hash out _onto each server_ referenced by a deep link. (The idea is that Android can use that file to "automatically verify" your deep link.)

On previous versions of Android, it was enough to just [register an intent](https://github.com/medic/cht-android/blob/fddf3afdf50606cc03410ad346b93cfe05b1c64f/src/main/AndroidManifest.xml#L66) with the configured `app_host` value. Once the app was installed on a device and the user clicked a link that included the `app_host` domain, they would be prompted to choose if they wanted to open the link in the app or in the browser.

image

**To Reproduce**

Now on Android 12+, once you install the app and open a link, you are never prompted to open it in the app (it just goes straight to the browser).  

**Expected behavior**

The only silver lining here is that once everything is configured for the deep links to actually work, the user should not even see the prompt for selecting where to open the link. Instead these auto-verified deep links should just open by default in the app. (Removing a point of confusion/error on the user side....)

**Environment**
- Android Version: (12+)

**Additional context**

You can actually manually re-enable the deep-link behavior on each device by manually verifying the links.  Open Settings > Apps > YOUR_APP (e.g. Medic (gamma)) > Open by default.  Select "Add link" and choose the configured link from the list.

sugat009:
This PR contains the changes for auto-verification of the domain with the app for deeplinks. This would only partially fix 308 .

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -59,7 +59,7 @@
     <activity android:name="AppUrlIntentActivity"
         android:launchMode="singleInstance"
         android:exported="true">
-      <intent-filter>
+      <intent-filter android:autoVerify="true">
         <action android:name="android.intent.action.VIEW" />
         <category android:name="android.intent.category.DEFAULT" />
         <category android:name="android.intent.category.BROWSABLE" />
--- a/src/moh_kenya_echis/AndroidManifest.xml
+++ b/src/moh_kenya_echis/AndroidManifest.xml
@@ -9,7 +9,7 @@
     <activity android:name="AppUrlIntentActivity"
       android:launchMode="singleInstance"
       android:exported="true">
-      <intent-filter>
+      <intent-filter android:autoVerify="true">
         <action android:name="android.intent.action.VIEW" />
         <category android:name="android.intent.category.DEFAULT" />
         <category android:name="android.intent.category.BROWSABLE" />


jkuester (approved):
:rocket: LGTM!  Just a reminder that we should document (in as much detail as possible) how we tested this functionality and what we learned about the behavior.  This can go either here in the PR or just in the issue thread.

jkuester:
@lorerod I removed you from this PR since there is not any change in functionality happening as a result of this (trying to keep unnecessary things off your plate!).   This functionality is currently non-operable and un-usable for the user because https://github.com/medic/cht-core/issues/8917 has not been addressed on the cht-core side.  The only reason for including this functionality here now in cht-android is because of the high program costs associated with deploying new apk versions.  We want to get this baked in now so it can start getting deployed, instead of waiting until we actually need it (which would be too late).  When we go to address the cht-core functionality, we will circle back for considerable workflow testing of this functionality.

sugat009:
> 🚀 LGTM! Just a reminder that we should document (in as much detail as possible) how we tested this functionality and what we learned about the behavior. This can go either here in the PR or just in the issue thread.

I was planning on writing the steps to test the changes from this PR in the issue itself because another PR for the manual verification and the e2e tests are incoming as well so I wanted to have the test process in one place for future references.

PR #356, opened 2024-04-10
Add moh civ apk generation and logo

inromualdo:


src/moh_civ_uat/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    package="org.medicmobile.webapp.mobile">
```
latin-panda, line 4:
```suggestion
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:tools="http://schemas.android.com/tools">
```
For example look at [moh-togo-uat](https://github.com/medic/cht-android/blob/master/src/moh_togo_uat/AndroidManifest.xml)

src/moh_civ_uat/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
...
+
+<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
```
latin-panda, line 6:
```suggestion
  <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
```
Indentation

latin-panda (changes requested):
A tiny fix please

latin-panda (approved):
LGTM!

Issue #326, opened 2023-08-24
Upgrade dependencies


latin-panda:
**Describe the issue**
The project's dependencies are falling far behind the latest versions. 

- Upgrade the `androidx`
- Upgrade the `com.android.tools`
- ~~Upgrade the testing dependencies, our current version doesn't support Target 33 (Android 13)~~
- ~~Enable the test files [here](https://github.com/medic/cht-android/pull/325) to run on Target 33.~~ taken care of in #349

PR #360, opened 2024-05-09
Upgrade `androidx` and Gradle dependencies

m5r:
#326

Code health PR to keep dependencies up-to-date

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,13 +106,16 @@ android {
     }
   }
 
+  base {
+    archivesName = "${project.name}-${getVersionName()}"
+  }
+
   defaultConfig {
     //noinspection OldTargetApi
     targetSdkVersion 34
     minSdkVersion 21 // Android 5.0
     versionCode getVersionCode()
     versionName getVersionName()
-    archivesBaseName = "${project.name}-${versionName}"
     //For espresso tests
     testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     // Test user credentials
--- a/build.gradle
+++ b/build.gradle
@@ -433,16 +433,16 @@ android {
 
 dependencies {
   implementation fileTree(dir: 'libs', include: ['*.jar'])
-  implementation platform('org.jetbrains.kotlin:kotlin-bom:1.9.23')
+  implementation platform('org.jetbrains.kotlin:kotlin-bom:1.9.24')
   implementation 'androidx.core:core:1.13.1'
   implementation 'androidx.activity:activity:1.9.0'
   implementation 'androidx.fragment:fragment:1.7.0'
-  compileOnly 'com.github.spotbugs:spotbugs-annotations:4.8.3'
+  compileOnly 'com.github.spotbugs:spotbugs-annotations:4.8.5'
   coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.4'
   testImplementation 'junit:junit:4.13.2'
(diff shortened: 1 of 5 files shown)


m5r:
@latin-panda could you take a look at this PR?

build.gradle
```diff
@@ -9,7 +9,7 @@ buildscript {
     google()
   }
   dependencies {
-    classpath 'com.android.tools.build:gradle:8.3.0'
+    classpath 'com.android.tools.build:gradle:8.4.0'
```
latin-panda, line 12: When I try to run it using the IDE I get this error:
`Minimum supported Gradle version is 8.6. Current version is 8.4.`

When using the `make` it works fine. Are there many changes if we upgrade to 8.6?

build.gradle
```diff
@@ -46,8 +46,10 @@ task pmd(type: Pmd) {
   exclude '**/**.mp3'
   exclude 'test/bash/bats/**'
   dependencies {
-    pmd 'net.sourceforge.pmd:pmd-java:6.55.0'
-    pmd 'net.sourceforge.pmd:pmd-xml:6.55.0'
+    pmd 'net.sourceforge.pmd:pmd-cli:7.1.0'
+    pmd 'net.sourceforge.pmd:pmd-ant:7.1.0'
```
latin-panda, line 50: Are these 2 required now? :O

src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java
```diff
@@ -209,7 +209,7 @@ public SettingsException(String message) {
 }
 
 class IllegalSettingsException extends SettingsException {
-
+	public static final long serialVersionUID = -6920159047512297868L;
```
latin-panda, line 212: Maybe worth adding a comment to explain this serial number

`// see https://docs.pmd-code.org/pmd-doc-6.36.0/pmd_rules_java_errorprone.html#missingserialversionuid`

And can you please update the link in L203 to `https://docs.pmd-code.org/pmd-doc-6.36.0/pmd_rules_java_errorprone.html#missingserialversionuid`?

latin-panda (commented):
Thanks Mokhtar! It's working fine but I just tested with Android 13 
I added some questions below

build.gradle
m5r: Not that I know of, it should "just work". I tried it again from scratch by re-cloning the repo and running it using the IDE and it worked fine.
It's probably specific to gradle <-> android studio, I'm not very familiar with it but could there be an integration to upgrade on your end? Have you tried running "Sync Project with Gradle Files"?

build.gradle
m5r: Yes they split up their code into different packages in v7.0 and these two are now required

src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java
m5r: Yes good point! I made both links point to the v7.1.0 of the docs since that's what we're using now
https://docs.pmd-code.org/pmd-doc-7.1.0/pmd_rules_java_errorprone.html#missingserialversionuid

build.gradle
latin-panda: I made a fresh install and now it's good, something dirty or old

latin-panda (approved):
LGTM! Thanks Mokhtar! ✨

PR #363, opened 2024-05-23
Adding flavor for lumbini.app

binokaryg:
APK was requested here:
https://github.com/medic/config-moh-nepal/issues/1293


jkuester (approved):
Double-checked everything against the guide and it all looks good!  LGTM!

Issue #338, opened 2023-11-12
Play Store reporting Medic app crashes


garethbowen:
**Describe the bug**
Play Store is reporting the "Medic" android app is crashing for 15% of users many with a `android.app.Fragment$InstantiationException`. The Medic app isn't "production" but we should investigate to ensure this doesn't impact anyone else, and prioritise accordingly.

**To Reproduce**
Not sure how to reproduce it.

**Expected behavior**
No crashes.

**Logs**
```
Exception java.lang.RuntimeException: Unable to start activity ComponentInfo{org.medicmobile.webapp.mobile/org.medicmobile.webapp.mobile.EmbeddedBrowserActivity}: android.app.Fragment$InstantiationException: Unable to instantiate fragment t.v: could not find Fragment constructor
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:3491)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:3643)
  at android.app.ActivityThread.handleRelaunchActivityInner (ActivityThread.java:5505)
  at android.app.ActivityThread.handleRelaunchActivity (ActivityThread.java:5404)
  at android.app.servertransaction.ActivityRelaunchItem.execute (ActivityRelaunchItem.java:69)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ClientTransactionHandler.executeTransaction (ClientTransactionHandler.java:58)
  at android.app.ActivityThread.handleRelaunchActivityLocally (ActivityThread.java:5464)
  at android.app.ActivityThread.access$3400 (ActivityThread.java:244)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2115)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:223)
  at android.app.ActivityThread.main (ActivityThread.java:7707)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:612)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:997)
Caused by android.app.Fragment$InstantiationException: Unable to instantiate fragment t.v: could not find Fragment constructor
  at android.app.Fragment.instantiate (Fragment.java:562)
[...]

PR #366, opened 2024-06-26
fix(#338): Refactor OpenSettingsDialogFragment class to have a empty constructor

sugat009:
Fixes: #338

Refactored the `OpenSettingsDialogFragment` class to have an empty constructor.
According to the Android documentation, every `Fragment` class should have an empty constructor that does nothing and is not a non-empty constructor. [Reference](https://developer.android.com/reference/android/app/Fragment.html#Fragment()). Not doing this leads to the Android ecosystem being unable to restore the Fragment's state when re-instantiated causing the issue mentioned in the linked issue.

Testing:
For e2e testing, you can start using the app normally and ensure that it does not crash and to test the functionality that the `OpenSettingsDialogFragment` class provides, you can tap the screen 5 times and then, right-swipe the screen with 2 fingers.

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
@@ -13,14 +13,15 @@ import androidx.annotation.Nullable;
 
 import java.time.Clock;
 
-@SuppressLint("ValidFragment")
 public class OpenSettingsDialogFragment extends Fragment {
 
-	private final View view;
 	private int fingerTapCount = 0;
 	private long lastTimeTap = 0;
 	private GestureHandler swipeGesture;
 	private static final int TIME_BETWEEN_TAPS = 500;
+
+	public OpenSettingsDialogFragment() {
+	}
 
 	private final OnTouchListener onTouchListener = new OnTouchListener() {
 		@SuppressLint("ClickableViewAccessibility")
@@ -32,13 +33,10 @@ public class OpenSettingsDialogFragment extends Fragment {
 		}
 	};
 
-	public OpenSettingsDialogFragment(View view) {
-		this.view = view;
-	}
-
 	@Override
 	public void onCreate(@Nullable Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
+		View view = getActivity().findViewById(R.id.wbvMain);
 		view.setOnTouchListener(onTouchListener);
 	}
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
```diff
@@ -13,15 +13,16 @@
...
-@SuppressLint("ValidFragment")
 public class OpenSettingsDialogFragment extends Fragment {
 
-	private final View view;
 	private int fingerTapCount = 0;
 	private long lastTimeTap = 0;
 	private GestureHandler swipeGesture;
 	private static final int TIME_BETWEEN_TAPS = 500;
 
+	public OpenSettingsDialogFragment() {
```
garethbowen, line 23: I think this is unnecessary as the default constructor will work, right?

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
```diff
@@ -32,13 +33,10 @@ public boolean onTouch(View view, MotionEvent event) {
...
 
-	public OpenSettingsDialogFragment(View view) {
-		this.view = view;
-	}
-
 	@Override
 	public void onCreate(@Nullable Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
+		View view = getActivity().findViewById(R.id.wbvMain);
 		view.setOnTouchListener(onTouchListener);
```
garethbowen, line 40: This solution looks so much cleaner! Have you tested that the touch listener still fires and works as expected?

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
sugat009: Yes, it does work without this. Removing this.

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
sugat009: Yes, tested this out on a physical device.

m5r: I also gave it a shot on an emulated device and it still works as expected :+1:

m5r (approved):
Looks good! One fewer warning in our logs :smile:

latin-panda (approved):
LGTM! Thanks Sugat!

tatilepizs (approved):
This looks good, thank you @sugat009. ⭐

Issue #373, opened 2024-09-17
Instrumentation tests CI job failing beacuse of deprecated `actions/upload-artifact` verison


jkuester:
The Instrumentation Tests job for our Build workflow [is failing](https://github.com/medic/cht-android/actions/runs/10827672121/job/30041362259)on GitHub Actions because the version of `actions/upload-artifact` is too old.  

https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

PR #374, opened 2024-09-17
chore(#373): update `actions/upload-artifact` action image to v4

jkuester:
Closes https://github.com/medic/cht-android/issues/373

I reviewed the [migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md), but our use case is so simple that I do not for-see any changes being necessary.

I confirmed the new version works as expected  by intentionally pushing a commit to this branch that broke a test.  You can see the uploaded artifact here: https://github.com/medic/cht-android/actions/runs/10906781402/artifacts/1943597746

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,7 +150,7 @@ jobs:
         script: make test-ui-gamma
 
     - name: Archive Results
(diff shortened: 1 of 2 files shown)


mrjones-plip (approved):
:qatar:

Issue #376, opened 2024-10-01
Google reporting app crashes on some phones


jonathanbataire:
**Describe the bug**
Google play is reporting app crashes on TECNO TECNO-KF6j (TECNO SPARK 7) & TECNO TECNO-KG5j (TECNO SPARK 8C)
```
Exception java.lang.RuntimeException:
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:3628)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:3873)
  at android.app.servertransaction.LaunchActivityItem.execute (LaunchActivityItem.java:85)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2197)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:268)
  at android.app.ActivityThread.main (ActivityThread.java:8016)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:627)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:997)
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.setOnTouchListener(android.view.View$OnTouchListener)' on a null object reference
  at L.H.onCreate (SourceFile:17)
  at android.app.Fragment.performCreate (Fragment.java:2486)
  at android.app.FragmentManagerImpl.moveToState (FragmentManager.java:1261)
  at android.app.FragmentManagerImpl.moveFragmentToExpectedState (FragmentManager.java:1581)
  at android.app.FragmentManagerImpl.moveToState (FragmentManager.java:1642)
  at android.app.FragmentManagerImpl.dispatchMoveToState (FragmentManager.java:3052)
  at android.app.FragmentManagerImpl.dispatchCreate (FragmentManager.java:2999)
  at android.app.FragmentController.dispatchCreate (FragmentController.java:173)
  at android.app.Activity.onCreate (Activity.java:1601)
  at org.medicmobile.webapp.mobile.EmbeddedBrowserActivity.onCreate (SourceFile:1)
  at android.app.Activity.performCreate (Activity.java:8077)
[...]

PR #377, opened 2024-10-04
fix(#376): fix issue of the app crashing in some phones

sugat009:
fixes: #376 

The issue seems to be a case when during some events the `OpenSettingsDialogFragment` is not able to find its calling activity which in this case is `EmbeddedBrowserActivity` and since it can't find `EmbeddedBrowserActivity` it's also not able to find the view `wbvMain` which in turn is unable to set the `setOnTouchListener` onto the view because the view is `null`. So, I've added a fallback and retry mechanism to give time for the `EmbeddedBrowserActivity` to load and be "findable" and also retain the instance of `OpenSettingsDialogFragment` because it's just a background listener that listens for taps and swipes.

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
@@ -22,7 +22,7 @@ public class OpenSettingsDialogFragment extends Fragment {
 	private GestureHandler swipeGesture;
 	private static final int TIME_BETWEEN_TAPS = 500;
 	private View mainView;
-	private boolean isViewSetup = false;
+	boolean isViewSetup = false;
 
 	private final OnTouchListener onTouchListener = new OnTouchListener() {
 		@SuppressLint("ClickableViewAccessibility")
@@ -62,13 +62,17 @@ public class OpenSettingsDialogFragment extends Fragment {
 		return false;
 	}
 
-	private void setupViewWithRetry() {
+	void setupViewWithRetry() {
 		new Handler(Looper.getMainLooper()).post(new Runnable() {
 			@Override
 			public void run() {
-				if (shouldSetupView() && !setupView()) {
-					new Handler(Looper.getMainLooper()).postDelayed(this, 100);
+				if (!shouldSetupView() || !setupView()) {
+					retrySetup();
 				}
+			}
+
+			private void retrySetup() {
+				new Handler(Looper.getMainLooper()).postDelayed(this, 100);
 			}
 		});
 	}
--- a/src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
@@ -5,6 +5,9 @@ import android.app.Activity;
 import android.app.Fragment;
 import android.content.Intent;
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.util.Log;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.View.OnTouchListener;
@@ -19,6 +22,8 @@ public class OpenSettingsDialogFragment extends Fragment {
 	private long lastTimeTap = 0;
 	private GestureHandler swipeGesture;
 	private static final int TIME_BETWEEN_TAPS = 500;
+	private View mainView;
+	private boolean isViewSetup = false;
 
 	private final OnTouchListener onTouchListener = new OnTouchListener() {
 		@SuppressLint("ClickableViewAccessibility")
@@ -33,8 +38,32 @@ public class OpenSettingsDialogFragment extends Fragment {
 	@Override
 	public void onCreate(@Nullable Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
-		View view = getActivity().findViewById(R.id.wbvMain);
-		view.setOnTouchListener(onTouchListener);
+		setRetainInstance(true);
+	}
+
+	@Override
+	public void onActivityCreated(Bundle savedInstanceState) {
+		super.onActivityCreated(savedInstanceState);
+		setupViewWithRetry();
+	}
+
+	private void setupViewWithRetry() {
+		new Handler(Looper.getMainLooper()).post(new Runnable() {
+			@Override
+			public void run() {
+				if (getActivity() != null && !isViewSetup) {
+					View view = getActivity().findViewById(R.id.wbvMain);
+
+					if (view != null) {
+						mainView = view;
+						mainView.setOnTouchListener(onTouchListener);
+						isViewSetup = true;
+					} else {
+						new Handler(Looper.getMainLooper()).postDelayed(this, 100);
+					}
+				}
+			}
+		});
 	}
 
 	private void countTaps(MotionEvent event) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
latin-panda, line 1: Is this code compatible with CHT's supported SDK (>=21) ?

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
```diff
@@ -59,6 +61,8 @@ public void setup() {
 		doNothing().when(openSettingsDialogFragment).startActivity(argsStartActivity.capture());
 
 		openSettingsDialogFragment.onCreate(null);
+		openSettingsDialogFragment.onActivityCreated(null);
+		shadowOf(Looper.getMainLooper()).idle();
```
latin-panda, line 65: Is this test already covering the error when the fragment is unavailable, and do the retry?

latin-panda (commented):
Just a couple of questions

How did you reproduce this error in phones? [Following Binod's steps](https://github.com/medic/cht-android/issues/376#issuecomment-2389659761)?

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
sugat009: yes, both `Looper` and `Handler` have been available since API 1 as per the Android docs.
https://developer.android.com/reference/android/os/Looper
https://developer.android.com/reference/android/os/Handler

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
sugat009: Not at the moment. I saw that the tests in this file were related to the behavior of the Fragment and decided not to add the test as it's an internal functionality. I could add it if it's needed.

latin-panda: Is it possible to add a unit test to catch this bug in particular? It can be a good addition to prevent regressions and also as an example for similar future implementations

latin-panda: The [setRetainInstance](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#setRetainInstance(boolean)) is deprecated in androidx, this unit test can help the developer to consider that case when migrating to a more recent androidx lib.

sugat009: Sure, let me add a test for that.

src/test/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragmentTest.java
```diff
@@ -1,232 +1,308 @@
...
+import org.robolectric.shadows.ShadowLooper;
 
 import java.time.Clock;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.util.stream.IntStream;
 
-@RunWith(RobolectricTestRunner.class)
+@RunWith(Enclosed.class)
+@SuppressWarnings("PMD.TestClassWithoutTestCases")
```
sugat009, line 45: @latin-panda I've created a nested class to enclose the testing of the `setupViewWithRetry` method because the mocks of the previous test cases were not reusable for it. Also, I've added this warning suppress statement because PMD was raising an error as to the outer class not having any test cases.

tatilepizs (approved):
Thank you for the fix @sugat009 ⭐ 

I was able to replicate the error using a Samsung Galaxy A01 phone with Android 10 on `master` branch.

[video](https://github.com/user-attachments/assets/a97e2d49-668d-46a4-88f9-2cb59f6e86e4)

And it was fixed on `376-google-reporting-app-crashes-on-some-phones` branch

[video](https://github.com/user-attachments/assets/60eade7a-0b80-4ff9-af50-bc6e05955f18)

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
```diff
@@ -33,8 +37,44 @@ public boolean onTouch(View view, MotionEvent event) {
 	@Override
 	public void onCreate(@Nullable Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
+		setRetainInstance(true);
```
jkuester, line 40: So, I am probably missing something here, but in my testing, adding this line here in the `onCreate` was all that I needed to avoid the error in question.  

In my debugging it seemed like without this line, the `OpenSettingsDialogFragment.onCreate` method was getting called a bunch of times when entering/exiting split-screen mode _before the `EmbeddedBrowserActivity.onCreate` method is even called_.  When this happened, the `wbvMain` view was still null (presumably because the activity was not created yet).  Eventually, the `EmbeddedBrowserActivity.onCreate` method is called for the new activity and that triggers the  `OpenSettingsDialogFragment.onCreate` method to be called again (and this time the `wbvMain` view exists.  

_However_, if I set `setRetainInstance(true);` here in `OpenSettingsDialogFragment.onCreate`  and leave everything else as it is in `master`, it prevents all the extra initial calls to  `EmbeddedBrowserActivity.onCreate` that were happening before `EmbeddedBrowserActivity.onCreate` and so avoids the error.  When the `EmbeddedBrowserActivity.onCreate`  function is called, that triggers a call to `OpenSettingsDialogFragment.onCreate` which then can access the `wbvMain` view at that point in the lifecycle.  

I am not 100% sure about exactly what is happening here. The  `EmbeddedBrowserActivity.onCreate` is not getting called multiple times. My theory is that there is some weird connection between the activity lifecycle and the fragment lifecycle that is causing the `OpenSettingsDialogFragment.onCreate` function to get called at other points during the activity lifecycle as the original `EmbeddedBrowserActivity` is destroyed.  Setting `setRetainInstance` seems to decouple the fragment from the activity lifecycle so that this does not happen.

jkuester (changes requested):

jonathanbataire:
Hey @sugat009 any update on this?

sugat009:
> Hey @sugat009 any update on this?

I've just started to continue looking into this.

src/main/java/org/medicmobile/webapp/mobile/OpenSettingsDialogFragment.java
sugat009: That's a strange one. I tried this one as well. My Android studio was crashing with the debug option. \o/. So, I did the classic debugging using logging. I did not get the `EmbeddedBrowserActivity.onCreate`  being called multiple times or `EmbeddedBrowserActivity.onCreate` not being called and only `OpenSettingsDialogFragment.onCreate` being called. I've included screenshots of this below.

Nevertheless, just adding the `setRetainInstance(true);` to line 38 on the master branch for ` OpenSettingsDialogFragment` class did work. (Tested on virtual and physical phones).

Setting the `setRetainInstance(true);` does detach the fragment from its activity which is perfect for this use case. [Docs](https://developer.android.com/reference/android/app/Fragment#setRetainInstance(boolean)).

I'll refactor the changes to include just the `setRetainInstance(true);` line in the `OpenSettingsDialogFragment`. I'll do this in a different branch and send another PR if this change is ever required in the future.

Screenshot_20241021_160923

Screenshot_20241021_161001

PR #380, opened 2024-10-23
chore: change release 1.4.1 to 1.5.0

sugat009:
Change release 1.4.1. in CHANGELOG to 1.5.0 because there seem to be other changes than #376

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 # Release notes
 
-## 1.4.1
+## 1.5.0
 
 ### Technical Improvements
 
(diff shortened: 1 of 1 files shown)


binokaryg:
> there seem to be other changes

Do we need to list them?

sugat009:
@binokaryg I've added in the other changes after 1.4.0, however, the others commits did not have much info on which ticket it belonged to or are from private repos.

binokaryg:
> @binokaryg I've added in the other changes after 1.4.0, however, the others commits did not have much info on which ticket it belonged to or are from private repos. Should I include the PR links instead?

yes, that can work

binokaryg (approved):
👍

PR #386, opened 2025-01-26
fix(#385): Production instance on production app should not have red border

### LINKED ISSUE #385: Red Border Displayed on Login Screen for Multi-Deployment Brands
### **Describe the Bug**

- https://github.com/medic/cht-android/issues/309

When using this feature to list multiple deployments to choose from, an issue occurs with deployments accessed via `*.app.medicmobile.org`. Upon selecting such a deployment, the app displays a red border on the login screen.

### **To Reproduce**
1. Add a valid deployment with a URL in the format `*.app.medicmobile.org` to the [instances list](https://github.com/medic/cht-android/blob/master/src/moh_kenya_echis/res/xml/instances.xml).
2. Build, open the app and select the deployment.
3. Wait for the login screen to load.
4. Observe the red border displayed around the login screen.

### **Expected Behavior**
The app should not display a red border on the login screen.
There should at least be a way to suppress this behavior for a selected brand.
There is `IS_TRAINING_APP` flag, which produces orange borders.
https://github.com/medic/cht-android/blob/1ce2a9cd73261d065437fdbd7450af8ef831d4a8/build.gradle#L426

Could we add a flag that suppresses any border? `IS_PRODUCTION_APP` ?

### **Screenshots**
Red Border Issue Screenshot

### **Environment**
- **Instance:** `*.app.medicmobile.org`
- **Android Version:** Any
- **App Version:** 1.5.0

**Additional context**
The issue is caused by the red border logic in `EmbeddedBrowserActivity.java`:

Code reference:
https://github.com/medic/cht-android/blob/4f2cc566b31bd10218403d934289e0f777295c05/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java#L89-L93

For multi-deployment brands, `settings.allowsConfiguration()` always returns true because they lack the `appHost` setting.

https://github.com/medic/cht-android/blob/4f2cc566b31bd10218403d934289e0f777295c05/src/main/java/org/medicmobile/webapp/mobile/SettingsStore.java#L95-L102

Single Deployment Brands: The `app_host` property is defined:

https://github.com/medic/cht-android/blob/4f2cc566b31bd10218403d934289e0f777295c05/src/musomali/res/values/strings.xml#L2-L5

Multi-Deployment Brands: The `app_host` property is missing, causing the app to treat them as unbranded:

https://github.com/medic/cht-android/blob/4f2cc566b31bd10218403d934289e0f777295c05/src/moh_kenya_echis/res/values/strings.xml#L2-L5

Previously, we used to have either:
- unbranded app with multiple instance URLs, or
- branded app with a single URL

That is not the case now with #309. Now we can have a branded app with multiple instances.

**Possible Solutions**
- Modify `allowsConfiguration()` in `SettingsStore` to account for multi-deployment brands that lack `appHost`.

- Introduce a dedicated setting for multi-deployment brands to ensure they are correctly branded.

- Adjust the logic in  EmbeddedBrowserActivity` to check branding status more reliably.

**Related Issues**
- https://github.com/medic/cht-android/issues/266
- https://github.com/medic/cht-android/issues/278

binokaryg:
Fixes #385 by:
- Adding a Boolean resource to indicate if it's a production instance
- Checking the resource to avoid displaying red border

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
@@ -19,7 +19,6 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
-import android.content.res.Resources;
 import android.net.ConnectivityManager;
 import android.net.Uri;
 import android.os.Bundle;
--- a/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
@@ -19,6 +19,7 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.content.res.Resources;
 import android.net.ConnectivityManager;
 import android.net.Uri;
 import android.os.Bundle;
@@ -80,13 +81,13 @@ public class EmbeddedBrowserActivity extends Activity {
 
 		this.settings = SettingsStore.in(this);
 		this.appUrl = settings.getAppUrl();
-
+		boolean isProductionApp =  getResources().getBoolean(R.bool.production);
 		this.requestWindowFeature(Window.FEATURE_NO_TITLE);
 		setContentView(R.layout.main);
 
 		// Add an alarming red border if using configurable (i.e. dev)
 		// app with a medic production server.
-		if (settings.allowsConfiguration() && appUrl != null && appUrl.contains("app.medicmobile.org")) {
+		if (settings.allowsConfiguration() && appUrl != null && appUrl.contains("app.medicmobile.org") && !isProductionApp) {
 			View webviewContainer = findViewById(R.id.lytWebView);
 			webviewContainer.setPadding(10, 10, 10, 10);
 			webviewContainer.setBackgroundResource(R.drawable.warning_background);
--- a/src/main/res/values/bools.xml
+++ b/src/main/res/values/bools.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <bool name="production">False</bool>
+  <bool name="production">false</bool>
 </resources>
--- a/src/main/res/values/bools.xml
+++ b/src/main/res/values/bools.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <bool name="production">False</bool>
+</resources>


mrjones-plip:
Excellent work tracking this down!  I really love how [well researched the parent ticket](https://github.com/medic/cht-android/issues/385) it is.

Does this PR allow [each flavor/brand of the APK](https://docs.communityhealthtoolkit.org/building/guides/android/branding/#2-new-brand) to have it's own `isProductionApp` value such that every time each APK is built it retains it's own production boolean value?

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
```diff
@@ -80,13 +80,13 @@ public void onReceiveValue(String result) {
...
 		this.appUrl = settings.getAppUrl();
-
+		boolean isProductionApp =  getResources().getBoolean(R.bool.production);
 		this.requestWindowFeature(Window.FEATURE_NO_TITLE);
 		setContentView(R.layout.main);
 
 		// Add an alarming red border if using configurable (i.e. dev)
 		// app with a medic production server.
-		if (settings.allowsConfiguration() && appUrl != null && appUrl.contains("app.medicmobile.org")) {
+		if (settings.allowsConfiguration() && appUrl != null && appUrl.contains("app.medicmobile.org") && !isProductionApp) {
```
jkuester, line 89: Looking at this and the rest of the context aroud this functionality, I wonder if the core issue is that we should have updated this line when [adding support for multiple prod instances](https://github.com/medic/cht-android/pull/313) so that it does not apply the border when `allowCustomHosts = False`.  Basically, I think setting this would get the behavior you need:

```suggestion
		if (settings.allowCustomHosts() && appUrl != null && appUrl.contains("app.medicmobile.org")) {
```

The goal of this red border (as it was originally implemented) was to warn when someone had entered a prod URL into the `Custom` url box of the Unbranded app.  When we did #313, we added a new `allowCustomHosts` setting to allow for toggling off the custom URL box for apps like `moh_kenya_echis` which presents a list of instances to choose from, but does not allow for entering a custom URL. As a part of that work, we should have changed this line from `allowsConfiguration` to `allowCustomHosts` because now `allowsConfiguration` just indicates if there is a list of multiple instances to choose from whereas it is  `allowCustomHosts` that would actually show the Custom URL box where a user _could_ enter a prod URL....

jkuester (changes requested):
Suggested a more direct approach that does not require a new property. Let me know if that makes sense and results in the desired behavior.

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
binokaryg: Disabling custom hosts would prevent us from leveraging this feature for production apps. While not a dealbreaker, it does limit the flexibility we could achieve with a single app. For example, it would restrict:

- Allowing users to log into arbitrary test instances for debugging or testing purposes.
- Enabling users to log into new production instances before those instances are formally integrated into the app.

mrjones-plip:
> Yes, you can override it with flavor specific value, similar to other values. If you don't provide, it will default to false, which is the current behavior.

Excellent! I see Josh has some more detailed input, but this was my only question - I'll defer to others on the rest of the PR - thanks!

src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java
jkuester: Okay, so maybe I am misunderstanding the scope of what you are trying to achieve here. Sorry!  I thought that the purpose of this PR was just to allow an app with the `moh_kenya_echis`-style multi-instance list to be able to include `app.medicmobile.org` URLs in the instance list without triggering the red border.

My understanding is that the goal of the red border, when it was introduced, was to be shown anytime someone entered a "production" URL into the Custom URL box (which at the time was only shown in the `Unbranded` app).  Now, it is technically possible to show the Custom URL option in other app flavors (by setting `allowCustomHosts = true` and not setting a dedicated `app_host`.  

@binokaryg do you have a new use-case now where you need to allow the Custom URL option in a production app and you do NOT want to see the red border (even if the user enters a "production" URL into the Custom box)? If that is the case, I think we may want to re-think the need for the red border at all, since it seems like the assumptions from the original purpose for the border are no longer holding true.... (And, frankly, using  `app.medicmobile.org` to determine if the URL is "production" is not even applicable for many self-hosted instances anyway.)

binokaryg: > anytime someone entered a "production" URL into the Custom URL

Is this a bad idea?

I'm asking this because that's how majority of the users are using it here.

> @binokaryg do you have a new use-case now where you need to allow the Custom URL option in a production app and you do NOT want to see the red border

Now we want to move to a branded app with a list of instances to choose from. Although it's not a requirement to have Custom URL option, it's still good to have for flexibility.

jkuester: I did some digging and it appears this functionality has been basically unchanged [since it was added](https://github.com/medic/cht-android/pull/19) 9 years ago.  That PR notes:

> Add an alarming red border if using configurable (i.e. dev) app with a medic production server.

Then, as now, the border got triggered when there was not a hard-coded `app_host` URL, but instead the user could enter a custom URL to connect to an instance.  

Based on the above quote, my inference is that folks were using the Unbranded (aka configurable) app to allow their users to connect to testing/training instances. The intention being to use the proper _branded_ app to connect to the actual production instance. Then, in order to prevent test/training data from being accidentally entered into a production instance, this border was added to warn the user that they were connecting to a production instance from an app intended for testing/training.

---

You mentioned:

> Is this a bad idea?
>
> I'm asking this because that's how majority of the app users are using it here.

Technically speaking, there should be no issue with connecting to a production instance via a Custom URL. (Assuming that is intended and there is no danger of test/training data being entered in prod). However, this is definitely not the recommended way to have users connect. Besides just having the extra step of making sure the prod URL is correctly entered _for each user_, you also will not be able to use any kind of [deep links](https://docs.communityhealthtoolkit.org/building/features/integrations/android/#sending-a-url) when connecting via a Custom URL.  

> Now we want to move to a branded app with a list of instances to choose from. 

My proposed solution of just changing `allowsConfiguration` to `allowCustomHosts` should make it so that the border is not shown when connecting to Medic prod apps chosen from a configured list.

> Although it's not a requirement to have Custom URL option, it's still good to have for flexibility.

This is where I think your approach here might differ from the previous thoughts/intentions that went into the red border.  That being said, if your existing users have already been connecting via a custom URL in the unbranded app and seeing the red border, then clearly the "warning" is really not providing much value (at least in your case).

In general, I think that the idea we can detect a "production" instance by checking the URL is now fundamentally flawed since there is no way to account for the URLs of self-hosted prod instances.  

## Suggested next steps

I think that the `allowsConfiguration` to `allowCustomHosts` change is the proper minimal fix for the  https://github.com/medic/cht-android/issues/385 bug and we should make that change to close the bug (either in this PR or I can open a different one).

For the case of your new branded app, I think [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) would argue that you should not enable the Custom URL box _until it is actually a requirement that you need it_.  

If you do identify a use case where it is necessary to support a Custom URL in your branded app (or use the Unbranded app in prod), I think we should open a new issue and the specifics of that use case will help direct the best way we can proceed with handling the red border.  It is tricky to anticipate potential needs and add logic to address possible cases. However, if/when we have a specific case, that can help us determine the best path forwards.  Several possible solutions would be:

- Remove the red border all together. Folks can use the Unbranded app and/or the Custom URL box at their own risk. The red boarder is imperfect anyway since it only flags Medic-hosted prod instances (and not self-hosted ones).
- If you need the Custom URL for connecting to test/training instances, you could instead publish a branded [training app](https://docs.communityhealthtoolkit.org/building/training/onboarding/#setting-up-a-training-app) with the links to the test instance(s) pre-configured.
- Add support for just disabling the red border for a particular brand (regardless of if the Custom URL box is shown)
- For a multi-instance brand we could add support for configuring the border to show/hide at the instance-by-instance case

jkuester (approved):
LGTM!

PR #388, opened 2025-01-29
Adding production flavor for MoH Nepal mHealth program

binokaryg:
For https://github.com/medic/config-moh-nepal/issues/1374


paulpascal (approved):
LGTM 🍞

PR #390, opened 2025-01-31
Case correction for flavor lumbini_chis

binokaryg:
For medic/config-lumbini#7

In the earlier PR (#389), the flavor case was not correct which resulted in failed build after adding tag.


paulpascal (approved):
Oupsss

PR #397, opened 2025-06-11
#537588 add options to load some URLs in the webview

### LINKED ISSUE #396: Support CHT SSO
**Is your feature request related to a problem? Please describe.**
Currently if you use the [new SSO feature](https://docs.communityhealthtoolkit.org/hosting/sso/) that shipped in `4.20` ([ticket](https://github.com/medic/cht-core/issues/9735)), when you click the “Login with SSO” button, you get sent to the default Android browser outside of the CHT APK webview.  While you can log in successfully, you end up being logged in in the default Android browser, and not in the CHT APK Webview.

**Describe the solution you'd like**
Enable users to log in with SSO inside the CHT Android APK

There might be a way to to get [the link verification](https://docs.communityhealthtoolkit.org/building/branding/android/#android-app-links-verification) to work such that when OIDC Provider redirected the user to the CHT in the browser, (see [step 7 on the sequence diagram](https://docs.communityhealthtoolkit.org/hosting/sso/technical/)), that it sends the user to the APK.  This is untested and likely won't work in a development environment

**Additional context**
* The SSO Feature works as expected in a browser on the desktop,  it's just the APK that doesn't work as expected
* There's a `isUrlRelated()` [function](https://github.com/medic/cht-android/blob/master/src/main/java/org/medicmobile/webapp/mobile/Utils.java#L30) that we might be able to extend to allow the OIDC provider URL to load in the webview
* This  `isUrlRelated()` function also might fetch the current OIDC URL dynamically so that every time you change the OIDC Provider, you don't need to ship a new manifest and build a new CHT Android APK
* Possibly related, the CHT has a [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) policy [defined](https://github.com/medic/cht-core/blob/master/api/src/routing.js#L200)

icrc-fdeniger:
add options to load some URLs in the webview
should fix #396

### DIFF
--- a/src/main/java/org/medicmobile/webapp/mobile/Utils.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/Utils.java
@@ -18,6 +18,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.File;
+import java.net.URI;
 import java.util.Map;
 import java.util.Optional;
 
@@ -31,7 +32,9 @@ final class Utils {
 		// android.net.Uri doesn't give us a host for URLs like blob:https://some-project.dev.medicmobile.org/abc-123
 		// so we might as well just regex the URL string
 		if (uriToTest != null) {
-			return isUrlRelated(appUrl, uriToTest.toString());
+			String uriToTestString = uriToTest.toString();
+			return isUrlRelated(appUrl, uriToTestString)
+					|| (uriToTestString.contains("redirect_uri="+ Uri.encode(appUrl)+Uri.encode("/medic/login/oidc")));
 		}
 		return false;
 	}
--- a/src/main/java/org/medicmobile/webapp/mobile/Utils.java
+++ b/src/main/java/org/medicmobile/webapp/mobile/Utils.java
@@ -18,7 +18,6 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.File;
-import java.net.URI;
 import java.util.Map;
 import java.util.Optional;
 
@@ -32,9 +31,7 @@ final class Utils {
 		// android.net.Uri doesn't give us a host for URLs like blob:https://some-project.dev.medicmobile.org/abc-123
 		// so we might as well just regex the URL string
 		if (uriToTest != null) {
-			String uriToTestString = uriToTest.toString();
-			return isUrlRelated(appUrl, uriToTestString)
-					|| (uriToTestString.contains("redirect_uri="+ Uri.encode(appUrl)+Uri.encode("/medic/login/oidc")));
+			return isUrlRelated(appUrl, uriToTest.toString());
 		}
 		return false;
 	}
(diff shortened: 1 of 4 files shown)


src/main/java/org/medicmobile/webapp/mobile/Utils.java
```diff
@@ -31,7 +32,9 @@ static boolean isUrlRelated(String appUrl, Uri uriToTest) {
 		// android.net.Uri doesn't give us a host for URLs like blob:https://some-project.dev.medicmobile.org/abc-123
 		// so we might as well just regex the URL string
 		if (uriToTest != null) {
-			return isUrlRelated(appUrl, uriToTest.toString());
+			String uriToTestString = uriToTest.toString();
+			return isUrlRelated(appUrl, uriToTestString)
+					|| (uriToTestString.contains("redirect_uri="+ Uri.encode(appUrl)+Uri.encode("/medic/login/oidc")));
```
jkuester: I am a bit worried about unintended consequences of including this logic down here in the generic `isUrlRelated` function. This function also gets called via `isValidNavigationUrl` [when the app is closed](https://github.com/medic/cht-android/blob/8020cd4924a0b92390d8c1543a3e9186c5e20c11/src/main/java/org/medicmobile/webapp/mobile/EmbeddedBrowserActivity.java#L161) (as part of the logic to save the user's current url to open next time).  If the app gets closed during the middle of an OIDC login flow, I don't think we want to save that URL.

jkuester (commented):
@icrc-fdeniger I have pushed a commit that moves the logic up to the `UrlHandler`. (See my inline comment for the reason.) Let me know if you see any problems with that and if that functionality still works on your end.  If all that is good, I think we can ship this! :+1:

jkuester:
@lorerod can you also give these changes a go and see if find any issues? :pray:

src/main/java/org/medicmobile/webapp/mobile/Utils.java
icrc-fdeniger: thanks @jkuester all good for me with your improvements.

lorerod:
Hi @jkuester, @icrc-fdeniger ,
Thank you for adding me to this PR and also for adding the unit tests. I’ve tested this branch locally by following the steps outlined in [these](https://docs.communityhealthtoolkit.org/hosting/sso/keycloak/) documentation, and everything is working as expected on my end. 

Here are videos of two tests conducted on my Galaxy A01 Core (Android 10):

<details><summary>Login with SSO when already logged in to Keycloak</summary>


https://github.com/user-attachments/assets/fb2e0d4e-84e7-4a84-bc25-8a3c2fcf5902


</details>

<details><summary>Login with SSO when not logged in to Keycloak</summary>


https://github.com/user-attachments/assets/3617b51a-41dc-4243-8179-66a22c14dcd1


</details>

This work looks good to me. Approving now.
Let me know if there’s anything else I should try or verify.

lorerod (approved):
Left [a comment ](https://github.com/medic/cht-android/pull/397#issuecomment-2989304386) with my test findings. This work looks great. Thank you @icrc-fdeniger.

PR #398, opened 2025-06-17
Tweaks needed for HealthPulse

jkuester:


src/main/java/org/medicmobile/webapp/mobile/ChtExternalApp.java
```diff
@@ -260,33 +265,36 @@ private JSONObject parseBundleToJson(Bundle bundle) {
...
-
 					json.put(key, jsonArray);
 					return;
 				}
 
 				Optional<Uri> imagePath = getImageUri(value);
 				if (imagePath.isPresent()) {
-					json.put(key, getImageFromStoragePath(imagePath.get()));
+					boolean keepFullResolution = bundle.getBoolean("sampleImage", false);
+					json.put(key, getImageFromStoragePath(imagePath.get(), keepFullResolution));
```
jkuester, line 297: This tweak allows us to save the returned image uncompressed when `samepleImage` is true. This is needed to be able to return the images marked for sampling to Audere (they cannot use compressed versions of the images when training the model).  

Note that the difference between `100` quality (full resolution) and `75` quality (compressed) may seem minor, but in my unscientific measurements I was seeing the ultimate size of the `png` file for the full resolution image to be `~5x` greater than the compressed image.  For example, when testing with my phone with a `16MP` camera, the compressed image size (of the png attachment uploaded to Couch) was `1.4MB`, but the uncompressed image size (saved when `sampleImage: true`) was `5.8MB`.

src/main/java/org/medicmobile/webapp/mobile/ChtExternalApp.java
```diff
@@ -260,26 +265,28 @@ private JSONObject parseBundleToJson(Bundle bundle) {
 			return null;
 		}
 
-		private void setValueInJson(String key, JSONObject json, Bundle bundle) {
+		private void setValueInJson(String parentKey, String childKey, JSONObject json, Bundle bundle) {
+			String key = ("detectedRdt".equals(parentKey) && "concerns".equals(childKey))
+				? "detectionConcerns"
+				: childKey;
```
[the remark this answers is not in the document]
jkuester: @Benmuiruri FTR, Audere has released a new version of the HealthPulse API, but they ended up shifting the naming to be a bit different from this.  Now the top-level `concerns` array is called `globalConcerns` while the nested `concerns` remains as `detectedRdt.concerns`.  We should not be able to use those names in our forms when loading output data from HealthPulse :+1:  (and this code here will no longer be necessary).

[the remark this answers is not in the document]
Benmuiruri: Thanks for the heads up.

src/main/java/org/medicmobile/webapp/mobile/ChtExternalAppHandler.java
```diff
@@ -21,13 +21,17 @@ public class ChtExternalAppHandler {
...
 
 	ChtExternalAppHandler(Activity context) {
 		this.context = context;
 	}
 
 	String processResult(int resultCode, Intent intent) {
 		try {
+			intent.putExtra("duration", (System.currentTimeMillis() - this.lastActivityStartTime) / 1000);
+			this.lastActivityStartTime = 0;
+
```
Benmuiruri: Are we including duration ?

src/main/java/org/medicmobile/webapp/mobile/ChtExternalApp.java
```diff
@@ -268,38 +269,56 @@ private JSONObject parseBundleToJson(Bundle bundle) {
...
+				// ODK/Enketo models a primitive multi-value list (e.g. a select_multiple question) as a
+				// space-delimited string.
+				String nodeList = primitiveListOpt
+					.stream()
+					.flatMap(List::stream)
+					.map(Object::toString)
+					.map(s -> s.replace(" ", "_"))
+					.collect(Collectors.joining(" "));
+				json.put(key, nodeList);
+				return;
```
Benmuiruri: From my understanding of the issue you opened in [cht-core](https://github.com/medic/cht-core/issues/10508) you wanted to move serialization of data out of cht-android. 

Isn't that this part ?

PR #423, opened 2026-01-22
fix(#422): use gradle v8.8

### LINKED ISSUE #422: bug: app bundling fails with current gradle version
### Description
As described [here](https://issuetracker.google.com/issues/402800800) versions >8.8 are bugged.
Builds with multiple apks enabled fail with https://github.com/medic/cht-android/actions/runs/21211501906/job/61027213421#step:45:221
```
Execution failed for task ':buildMoh_kenya_echisReleasePreBundle'.
> Sequence contains more than one matching element.
```

jonathanbataire:
fixes #422

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     google()
   }
   dependencies {
-    classpath 'com.android.tools.build:gradle:8.10.1'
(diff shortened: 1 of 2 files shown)


jkuester (approved):
😞 Someday we need to get Gradle upgraded, but that can be a problem for another day....

